PR #1279, opened 2023-08-11
Fix bug in `Country_Map_Raw`.

samussiah:
`Country_Map_Raw` wasn't actually using its argument - my bad!


mattroumaya (approved):
Good catch!!

PR #1304, opened 2023-09-08
reworded Overview table filter

### LINKED ISSUE #1278: Reword Overview Table Filter to be more accurate
Nitpick (fine to deal with in future release): 

image

Let's reword filter to "Sites with {Red|Red or Amber|Any} Flags". Want to avoid calling the sites themselves Red/Amber, etc.

_Originally posted by @jwildfire in https://github.com/Gilead-BioStats/gsm/pull/1242#discussion_r1291402088_

### LINKED ISSUE #1289: Bugfix: Remove `site_num` hard code in `Overview_Table`.
## Expected Behavior
Use mappings.

## Current Behavior
If `dfSITE$strSiteCol != 'site_num'` the site metadata can't be merged onto the overview table.

## Steps to Reproduce

1.
2.
3.
4.

andersonjake1988:
fix #1278 and fix #1289


R/Overview_Table.R
```diff
@@ -322,7 +321,8 @@ Overview_Table <- function(
         caption = HTML(overview_table_flagged_caption),
         options = list(
           language = list(
-            lengthMenu = paste0("Showing _MENU_ ", table_dropdown_label)
+            lengthMenu = paste0(if(strReportType == "site"){"Sites Containing _MENU_ "}
+                                else if(strReportType == "country"){"Countries Containing _MENU_ "}, table_dropdown_label)
```
mattroumaya, line 325:
```suggestion
            lengthMenu = paste0(if(strReportType == "site"){"Sites Containing _MENU_ "}
                                else if(strReportType == "country"){"View _MENU_  Countries"}, table_dropdown_label)
```

R/Overview_Table.R
```diff
@@ -39,20 +39,19 @@ Overview_Table <- function(
...
   if (strReportType == "site") {
     grep_value <- "kri"
-    table_dropdown_label <- "Sites"
+    table_dropdown_label <- "Flags"
   }
 
   if (strReportType == "country") {
     grep_value <- "cou"
-    table_dropdown_label <- "Countries"
+    table_dropdown_label <- "Flags"
```
mattroumaya, line 47:
```suggestion
    table_dropdown_label <- NULL
```

mattroumaya (approved):
Looking good @andersonjake1988! 

Left a minor comment to address the wording. 

In the country report, we don't use the conditional filter since it's not likely that there will be more than 10-20 countries (I think...)

So for the country report, we can default to *Show X Countries*. 

Once this change is made (or feel free to suggest better wording), feel free to merge.

R/Overview_Table.R
```diff
@@ -69,7 +68,7 @@ Overview_Table <- function(
       ) %>%
       left_join(
         dfSite %>%
-          mutate(site_num = as.character(.data$site_num)) %>%
+          mutate(site_num = as.character(.data[[Read_Mapping()$dfSITE$strSiteCol]])) %>%
```
samussiah, line 71: @andersonjake1988, @mattroumaya - I don't think this is going to work with non-standard column mappings.

andersonjake1988: well shoot... could you expand on non-standard column mapping? as far as what to expect, so I can modify this?

samussiah: Sure thing!  The input data to `{gsm}` might not have a column in `dfSITE` named `site_num`.  In this scenario the user/study team would modify the `dfSITE` column mapping as needed (`dfSITE$strSiteCol`).

Rather than pass the custom column mapping into `Overview_Table` we can just assume the mapped site data, the output of `Site_Map_Raw()`, contains a column named `siteid` ([pertinent code](https://github.com/Gilead-BioStats/gsm/blob/dev/R/Site_Map_Raw.R#L19-L21)).  `Site_Map_Raw()` adds a column named `siteid` by setting it to `dfSITE[[ mapping$dfSITE$strSiteCol ]]`.

samussiah: And to elaborate a bit more, we either want to work with raw data + mappings or with mapped data, so we're not making assumptions about the columns in the data.

PR #1402, opened 2023-11-09
Fix 1392 - Managing `clindata` as a Suggests dependency

### LINKED ISSUE #1392: Bugfix: Add remote dependency on `clindata`
Hi `gsm` team. Thanks for all your work on this package 🥇. At Roche we're currently assessing this package for possible internal use, and as part of that effort we're running it through our package review tools. 

We had some trouble reproducing the test suite due to a missing package `clindata` and I just wanted to propose a fix to make your dependency use more consistent with general package behaviors. The goal is that the `clindata` package is properly picked up as a remote dependency and its use in the package code doesn't result in errors when the package isn't installed. 

## Expected Behavior

Development dependencies can be installed without separate installation process (using `remotes::install_dev_deps` or `pak:: local_install_dev_deps`) and the lack of installed suggests dependencies does not introduce errors. 

## Current Behavior

* `clindata` will not be found as it's only available as a git repository. It must be installed separately.
* When not installed, some functions will error instead of having a fallback behavior when the Suggests package is unavailable.

## Possible Solution

To use `clindata` as a Suggests package (I think this should be the preferred approach, since data packages can be a bit heavy sometimes), it should be listed as a Remote and uses of the package should be protected by tests for `requireNamespace`.

This can be easily remedied by adding a `Remotes` dependency on `clindata`

`DESCRIPTION`
```diff
+ Remotes:
+   clindata=Gilead-BioStats/clindata
```

And checking for an installed version of `clindata` before using it in the body of the package. For example, you could include the following before attempting to use any `clindata` functions or objects:

```r
if (!requireNamespace("clindata", quietly = TRUE)) {
    stop("Package \"clindata\" must be installed to use this function.", call. = FALSE)
}
```

Alternatively, you could make `clindata` a required dependency by moving it to `Imports`, though I think this type of dependency is best suited for `Suggests`.

## Steps to Reproduce

0. (if needed) remove `clindata` using `remove.packages("clindata")`
1. Clone repo and set working directory to the to the source code directory root
2. Attempt to install development package using either `remotes::install_dev_deps(".")` or `pak:: local_install_dev_deps(".")`

mattroumaya:
## Overview
Fix #1392 
⚠️ Draft PR that is targeting `main` - do not merge. 

Showing a quick idea for managing `clindata` as a default function argument for `*_Map()` functions.

## Test Notes/Sample Code

With `clindata` available:

Check out this branch and `devtools::load_all()`. 

```
UseClindata(
  list(
    "dfSUBJ" = "clindata::rawplus_dm",
    "dfAE" = "clindata::rawplus_ae"
  )
)
```

Remove `clindata`:

```
remove.packages("clindata")

# restart R session

devtools::load_all()

UseClindata(
  list(
    "dfSUBJ" = "clindata::rawplus_dm",
    "dfAE" = "clindata::rawplus_ae"
  )
)

Error in `UseClindata()`:
! Package clindata must be installed to use this function with sample data. Run `remotes::install_github('Gilead-BioStats/clindata')`
Run `rlang::last_trace()` to see where the error occurred.
```


mattroumaya:
@jwildfire @dgkf-roche - here's an idea for #1392, and would take a similar approach/use similar helper functions for use [within `Study_Assess()`](https://github.com/Gilead-BioStats/gsm/blob/dev/R/Study_Assess.R#L49-L64) and other functions. 

Feel free to take a look and let me know if this seems like an acceptable solution. Thanks!

mattroumaya:
@dgkf-roche are you able to check this one out and let us know if this resolves the `{clindata}` issues you were seeing?

dgkf-roche:
> @dgkf-roche are you able to check this one out and let us know if this resolves the `{clindata}` issues you were seeing?

Tested it out and it's looking good to me! Thanks for your willingness to address it 👍

R/util-clindata.R
```diff
@@ -0,0 +1,35 @@
...
+#' @importFrom cli cli_abort
+#' @importFrom purrr map
+#' @export
+UseClindata <- function(lDomains = NULL) {
+  if (!requireNamespace("clindata", quietly = TRUE)) {
+    cli::cli_abort(
+      "Package {.pkg clindata} must be installed to use this function with sample data.\nRun {.code remotes::install_github('Gilead-BioStats/clindata')}"
+    )
+  } else {
+    clindata_list <- purrr::map(lDomains, ~ eval(parse(text = .x)))
```
samussiah, line 31: Clever!

samussiah (approved):
Love the approach!

PR #1418, opened 2023-11-24
Fix 1411

### LINKED ISSUE #1411: QC: Evaluate which tables should be retained in `Make_Snapshot()`.
## QC Details
- we should probably drop all of the old tables, but evaluate differences. 
- specifically, look at old tables and see if there is any data that is not retained in new tables.

andersonjake1988:
## Overview
fix #1411 

## Test Notes/Sample Code

Notes:


jwildfire (commented):
Looks good to me in general. Will leave formal approval and merge to @mattroumaya

.github/workflows/pkgdown.yaml
mattroumaya, line 1: Before we release `v1.9.0` we should check that the GH Actions reporting that adds a sample report to the pkgdown site works as intended.

mattroumaya (approved):

PR #1436, opened 2023-12-07
created function for converting past snapshots to v1.9.0 format

### LINKED ISSUE #1434: Create backwards compatibility conversion function
Create function to convert old data model saved snapshot data to the new data model

andersonjake1988:
## Overview
fix #1434 

## Test Notes/Sample Code

Notes:


R/util-MakeSnapshotHelpers.R
```diff
@@ -548,6 +547,7 @@ MakeRptThresholdParam <- function(meta_param, status_param, gsm_analysis_date, t
           !is.na(index) & !is.na(value) ~ value
         ),
         "studyid" = unique(status_param$studyid),
+        "snapshot_date" = {if(is.null(gsm_analysis_date)) "gsm_analysis_date" else gsm_analysis_date},
```
mattroumaya, line 550: there's a stray `"snapshot_date" = gsm_analysis_date` above

R/util-MakeSnapshotHelpers.R
```diff
@@ -410,17 +409,17 @@ MakeRptSiteKriDetails <- function(lResults, status_site, meta_workflow, meta_par
...
 #'
 #' @keywords internal
-MakeRptKriBoundsDetails <- function(lResults, config_workflow, gsm_analysis_date) {
+MakeRptKriBoundsDetails <- function(lResults, config_workflow, gsm_analysis_date = NULL) {
   bounds <- MakeResultsBounds(lResults = lResults, dfConfigWorkflow = config_workflow)
   if (length(bounds) > 0) {
     bounds %>%
       mutate(
-        "snapshot_date" = gsm_analysis_date,
+        "snapshot_date" = {if(is.null(gsm_analysis_date)) "gsm_analysis_date" else gsm_analysis_date},
```
mattroumaya, line 422: Is there a reason to favor using `"gsm_analysis_date"` over `NA`/`NA_Date_`?

I think I'd prefer changing these to something like:

`"snapshot_date" = ifelse(is.null(gsm_analysis_date), NA, gsm_analysis_date)`

mattroumaya (approved):
Looks great!! 

Approving with some comments for consideration - feel free to merge with or without changes.

R/util-MakeSnapshotHelpers.R
andersonjake1988: I had made it this was as to assume that if gsm_analysis_date is missing from the function call then It assumes that column is in the data (lStackedSnapshots should already have gsm_analysis_date. and snapshot date would then be defined by that column not the character "gsm_analysis_date") otherwise if it's missing it would just return the character representation. This change was only implemented to convert a past lStackedSnapshots (calculated with Augment_Snapshot) without altering past dates

PR #1442, opened 2023-12-13
Fix 1435

### LINKED ISSUE #1435: Review Reporting functions and verify they operate as intended with the new data model

andersonjake1988:
## Overview
fix #1435 

## Test Notes/Sample Code

Notes: 
QTL reports still in work, everything is generated except the charts, when the charting workflow is finalized I can make the necessary adjustments


mattroumaya (approved):
Nice! less involved than I thought this would be. 

going to merge for now so that we can resolve with the charts PR.

PR #1454, opened 2024-01-02
Fix 1426

### LINKED ISSUE #1426: Feature: Add `tryCatch` around each workflow in `Study_Assess` so all workflows run.
## Feature Details
`Study_Assess` fails silently when redirecting output to a log file. It would be helpful to allow for workflow failures without halting the overall process.

## Possible Implementation
```
walk(
  workflows,
  ~ {
    tryCatch(
      {
        runWorkflow()
      },
      warning = function(warning) {
        capture_warnings()
      },
      error = function(error) {
        capture_errors()
      },
      finally = {
        capture_workflow_status()
      }
  }
)
```

mattroumaya:
## Overview
Fix #1426 

Uses `purrr::safely()` in favor of `tryCatch()` to always run a workflow. Returns errors if there are any as `lAssessments$kriID$error`.


R/Study_Assess.R
```diff
@@ -104,5 +106,15 @@ Study_Assess <- function(
     lAssessments <- NULL
   }
 
+  # extract results
+  lAssessments <- purrr::imap(lAssessments, function(x, y) {
+    if (is.null(x$error)) {
+      x$result
+    } else {
+      if (!bQuiet) cli::cli_alert_warning("Error in workflow {y}: {x$error}")
```
andersonjake1988, line 114: just making sure this if statement is intentional. in the output right now it will return NULL if the the workflow had an error, unless bQuiet is TRUE then it will output the error code for the workflow instead of NULL

andersonjake1988 (commented):
Looks good Matt, just had one question about the output, and you needed to add safely to the imports, but other than that it functions as intended

R/Study_Assess.R
mattroumaya, line 114:
```suggestion
    } else {
      x$error
      if (!bQuiet) cli::cli_alert_warning("Error in workflow {y}: {x$error}")
```

R/Study_Assess.R
```diff
@@ -104,5 +106,22 @@ Study_Assess <- function(
...
+  lAssessments <- purrr::map(lAssessments, function(x) {
+    purrr::discard(x, is.null) %>%
+      purrr::flatten()
+  })
+
+  if (!bQuiet) {
+    purrr::iwalk(lAssessments, function(x, y) {
+
+      if ("message" %in% names(x)) {
+        cli::cli_alert_warning("Workflow {y} did not run. Error message: {x$parent$message}")
```
samussiah, line 119: `x$parent$message` doesn't appear to catch the error message emitted on workflow failure.

samussiah (changes requested):
Nice and to the point - love it @andersonjake1988!

I'm not seeing the error message captured though.  For instance, I just added `stopifnot('this is an error message' = FALSE)` to an assessment function and it's not being spit out at the end of `Study_Assess`.  If it's not a quick fix to capture that message perhaps just drop `Error message: {x$parent$message}` from `Study_Assess`.

R/Study_Assess.R
mattroumaya: @samussiah - thanks for reviewing! 

I want to test this out in the same way you're looking at it. Did you put that `stopifnot()` somewhere like [here](https://github.com/Gilead-BioStats/gsm/blob/dev/R/AE_Assess.R#L96)?

`purrr::safely()` captures the entire callstack in a nested list, so we could maybe parse some of that out if needed, but want to know what the most straightforward way to resolve the pipeline error issue would be. E.g, if we don't need to print the error to the console and can capture it further downstream, then we can just remove [this part](https://github.com/Gilead-BioStats/gsm/blob/543507289c34dd4824ee23c95e80b41fe67dab1d/R/Study_Assess.R#L115-L123).

samussiah: @mattroumaya I just threw a random (unexpected) error as opposed to an expected error like the ones in `stopifnot`. The console output will display unexpected errors that actually bomb the workflow, so I don't think it's worth the time to parse the call stack as long as the workflow exits gracefully.

samussiah: That said, it would be pretty slick...

andersonjake1988 (approved):

PR #1469, opened 2024-01-17
Fix `Timeline()` warnings during report generation

### LINKED ISSUE #1467: Bugfix: Suppress warnings from `Timeline()` in `inst/report/KRIReportBySite.Rmd`.
## Expected Behavior
Only the timeline itself appears in the report.

## Current Behavior
image

## Possible Solution
Suppress warnings or program around missing values.

## Steps to Reproduce

1.
2.
3.
4.

mattroumaya:
## Overview
Fix #1467 

Hoping that this easy change will suppress the warnings/messages.

### DIFF
--- a/inst/report/KRIReportByCountry.Rmd
+++ b/inst/report/KRIReportByCountry.Rmd
@@ -16,6 +16,10 @@ params:
     lCharts: NULL
 ---
 
+```{r setup, include=FALSE} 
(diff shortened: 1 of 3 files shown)


samussiah (approved):
Works for me - the issue was the first and/or last element in `breaks` not displaying in the timeline due to truncation (?).  Anyway, this fixes it.

PR #1482, opened 2024-02-02
added logic to use reports in local RStudio

### LINKED ISSUE #1481: Study_Report not rendering graphs and site-select widget not working
## The problem
When generating reports the non of the charts are rendering, and the site select widget has no options to select sites, Investigating the cause

andersonjake1988:
## Overview
fix #1481

## Test Notes/Sample Code
added logic to allow for report rendering in local RStudio

Notes:


mattroumaya (approved):
Nice - this looks like a safer option!

PR #1510, opened 2024-03-19
Fix 1484

### LINKED ISSUE #1484: Feature: Add version of package in qualification report

mattroumaya:
Fix #1484

### DIFF
--- a/.github/workflows/R-CMD-check-dev.yaml
+++ b/.github/workflows/R-CMD-check-dev.yaml
@@ -42,10 +42,11 @@ jobs:
       - name: install packages
         shell: Rscript {0}
         run: |
-          install.packages("devtools")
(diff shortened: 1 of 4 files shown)


samussiah (approved):
Nice and to-the-point!  Why does `{arrow}` need to be explicitly installed?

mattroumaya:
> Nice and to-the-point! Why does `{arrow}` need to be explicitly installed?

It looked like there was an issue that pops up at times with installing C++ dependencies on MacOS, and the simplest solution I could find was installing with `source = TRUE`. Not totally sure if this is needed, but this is the only time I can remember tests failing only on one OS and not others - [GH Actions log here](https://github.com/Gilead-BioStats/gsm/actions/runs/8350847108/job/23361080664)

Issue #1078, opened 2023-02-22
Feature: Identify Newly Flagged KRIs


mattroumaya:
## Feature Details
- When longitudinal data is available, identify sites with new red flags (+/- 2)
- Add boxes to existing report to show # of newly flagged sites per KRI
- Surface KRI history in study overview table -- display change from previous snapshot regardless of current KRI status.

Boxes should have:
-  n new red flags
-  n total red flags
-  n total amber flags
- some indication for each site that is newly flagged
**Likely to be done as part of update in `{rbm-viz}`**

## additional supplemental table at top of report
- just shows trends
- site number, basic info, filterable version of `{grail}` table
  - filter on: current KRI value, previous KRI value
  - e.g., prev KRI value = 0, current KRI value = +2, only show sites with that criteria

mattroumaya:
link to rbm-viz issue

mattroumaya:
@jwildfire to add details

jwildfire:
Here are my current thoughts about how to incorporate information about newly flagged KRIs in our standard KRI report in {gsm}. I think this data is all fairly easily available in the v1.9 data model. Once we agree on details, this should almost definitely be broken out in to multiple issues. 
 
**Overview Section**
- ** Show “XX Newly Flagged KRIs” next to “XX Red/Amber KRIs”
- ! Overview table - Add “Sites with New Flags” as option to filter 
- ! Overview table - Invert color scheme (Change background color to red/amber, with white icon) for newly flagged KRIs 
- ** Add a new listing of Newly flagged KRIs
    - Columns = Site, KRI, current_metric, current_score, current_flag, previous_metric, previous_score, previous_flag

**KRI charts/table**
- ! Scatter plot - use fill color to indicate newly flagged KRIs. Filled = new flag. Empty = no change. 
- ** Add Prev Flag to summary table
- Bar Charts - Add a line showing the change from the previous assessment
    - Probably overkill and don’t want to do this
- ** Add new heatmap grid showing all flag values. Columns = snapshots. Rows = sites. Fill color = flag status. 
    - Include filter for “Sites with Red/Amber” and all sites
    - Add a new chart tab for this
- ** Summary table. Add column for previous flag

! = rbm-viz update
** = target for v1.10

jwildfire:
cc: @Craftech-RR

PR #1570, opened 2024-06-08
Flags over time table

jwildfire:
## Overview

POC of a little heatmap showing flags over time. Related to #1078.

image

Next step is to convert to a widget add some basic functionality: 

- Mouseover for details​
- Controls
   - Show/hide score​
   -Show Columns​
      - Last 5/All ​
    - Show Rows:​
       - All​
       - New Flag​
       - Any Flag​

## Test Notes/Sample Code

# Test FlagOverTime chart
FlagOverTime(dfSummary, dfSite, dfMetrics)

```
dfSummary <- data.frame(
    GroupID = c(100, 100, 200, 200),
    GroupLevel = c("Site", "Site", "Site", "Site"),
    MetricID = c('kri1', 'kri1', 'kri1', 'kri1'),
    snapshot_date = c("2021-01-01", "2021-02-01", "2021-01-01", "2021-02-01"),
    Flag = c(-2, 0, 1, -1)
)

dfSite <- data.frame(
    GroupID = c(100, 200)
)

dfMetrics <- data.frame(
    MetricID = c('kri1', 'kri2'),
    abbreviation = c("AE", "SAE")
)

FlagOverTime(dfSummary, dfSite, dfMetrics)
```


jwildfire:
@jonthegeek take a look at this

jonthegeek:
@jwildfire Some questions:

- Looking at other function names, should this be `Widget_FlagOverTime()`? Or maybe `Report_FlagOverTime()`? Or both?
- Should we export a separate function for pivoting the flagged table toward dates (basically the first half of this function right now)? It makes sense but I'm not sure we have anything exactly equivalent. It isn't REALLY like the other `Flag_*()` functions, but that's probably the closest match (so it would be something like `Flag_Longitudinal()` or `Flag_OverTime()`)
- Should this be added to any existing report(s)?
- Is there good sample data handy that would show what we're trying to show? The data above is a little too simple for me to trust that I'm seeing the "real" thing. Maybe something to be added to #1629?
- What details should show on hover?

jwildfire:
> * Looking at other function names, should this be `Widget_FlagOverTime()`? Or maybe `Report_FlagOverTime()`? Or both?

I think `Report_FlagOverTime()` is probably ok unless we decide to turn it into an actual `htmlwidget`. 

> * Should we export a separate function for pivoting the flagged table toward dates (basically the first half of this function right now)? It makes sense but I'm not sure we have anything exactly equivalent. It isn't REALLY like the other `Flag_*()` functions, but that's probably the closest match (so it would be something like `Flag_Longitudinal()` or `Flag_OverTime()`)

I'm leaning towards just keeping the pivot in the reporting function for now since it's currently only used in this one place. If we end up needing it in other functions in the future we can pull it out. 

> * Should this be added to any existing report(s)?

Yes, I'd like to add it to the KRI report once it's ready. 

> * Is there good sample data handy that would show what we're trying to show? The data above is a little too simple for me to trust that I'm seeing the "real" thing. Maybe something to be added to [Add sample reporting data #1629](https://github.com/Gilead-BioStats/gsm/pull/1629)?

Yeah, we need a good longitudinal data dfSummary df as part of #1629. I wrote a little simulation [here](inst/examples/LongitudinalReport.R)  that we could use to generate potentially. 

> * What details should show on hover?
I was thinking `Numeratror`, `Denominator`, `Metric` and `Score` from dfSummary, but it's really just a nice-to-have not a requirement

jonthegeek:
@jwildfire Does this work for this PR? It isn't actually USED anywhere at this point, but I want to check in and maybe split off a separate PR for the report integration.

jonthegeek:
Current example:

FlagOverTime

jonthegeek:
Adding hovers (for full date and more data about the flag cells) will take a considerable amount of more work, so I'm saving that for a future PR. Otherwise this seems pretty good already. Pagination and filtering will definitely help, but this is usable at this point.

jonthegeek:
Converted back to draft pending merge of #1629. We'll use that summary data here.

jonthegeek:
TODO:

- [ ] Test locally with real data.
- [ ] Investigate making dfMetrics optional.

jonthegeek:
@jwildfire This works for the multi-year test dataset, so I think this is good to go. It might change once things settle down with regards to dfSummary vs dfResults, but as of today I think it's following the patterns of other functions.

I split out some gt utility functions, notably `gsm_gt()`, which is intended to replace `gt::gt()` but also apply any styling rules we agree upon (mostly because you happened to include a style in the stub of this function).

Let me know if you'd like to change anything!

Issue #1543, opened 2024-05-10
Feature: Add control to highlight all sites from a country in `widget_scatterplot`


jwildfire:
## Feature Details

Once country highlight [rbm-viz feature](https://github.com/Gilead-BioStats/rbm-viz/issues/372) is implemented we need to add a new dropdown in `widget_scatterplot` that allows a user to select a country to highlight. 

## Example Code
This update applies to [`inst/htmlwidgets/Widget_ScatterPlot.js`]

- Add a dropdown with a list of countries out of `dfSite`: `country`
- define `addCountrySelect` with an event listener that passes the site IDs back to the chart

## Possible Implementation
![image](https://github.com/Gilead-BioStats/gsm/assets/5428548/a5d3e631-e64f-4c45-b9c1-00a7a42cb5a2)

jwildfire:
@samussiah - could @taylorrodgers work on this once you've got a rbm-viz branch?

samussiah:
```
// Add event listener to highlight countries.
const country = function (datasets, setup = false) {
    let instance = getChart();
    const countryDropdown = document.querySelector('#country');

    if (setup) {
        const option = document.createElement('option');
        option.value = 'None';
        option.innerHTML = 'None';
        countryDropdown.appendChild(option);

        const countries = [...new Set(datasets[3].map((d) => d.country))].sort(
            (a, b) => a - b
        );

        for (const i in countries) {
            const option = document.createElement('option');
            option.value = countries[i];
            option.innerHTML = countries[i];
            countryDropdown.appendChild(option);
        }

        countryDropdown.value = 'None';

        countryDropdown.addEventListener('change', (event) => {
            instance = getChart();
            instance.data.config.selectedGroupIDs = datasets[3]
                .filter(d => d.country === event.target.value)
                .map(d => d.siteid);
            instance.data.config.xType = xAxisType();
            instance.helpers.updateConfig(instance, instance.data.config);
        });
    }

    return countryDropdown.value;
};
```

samussiah:
@taylorrodgers - this is a good feature to tackle ahead of {gsm}@v2.  Add a dropdown of countries that highlights sites in the selected country, something that adds a control to the widget like [`addSiteSelect`](https://github.com/Gilead-BioStats/gsm/blob/dev/inst/utils-0.0.1/addSiteSelect.js) and passes an array of sites to the chart.

PR #1619, opened 2024-07-02
Add country dropdown to scatterplot

taylorrodgers:
## Overview
This was for #1543. I created `addCountrySelect.js` to `inst/utils`. This required that we modify `inst/htmlwidgets/Widget_ScatterPlot.js` and `inst/htmlwidgets/Widget_ScatterPlot.yaml` to call the `addCountrySelect.js` function.

## Test Notes/Sample Code
To test, run the following after running lines `1-97` on  `workflow_reporting_charts.R`.

You can see the functionality below:

https://github.com/Gilead-BioStats/gsm/assets/76186904/f6c96e49-8d9b-4704-b2bf-5f7daa6c3003


inst/utils-0.0.1/addCountrySelect.js
```diff
@@ -0,0 +1,77 @@
...
+    countrySelectContainer.appendChild(countrySelect);
+
+    // add default option
+    const noneOption = document.createElement('option');
+    noneOption.innerHTML = 'None';
+    countrySelect.appendChild(noneOption);
+
+    // get sorted array of country
+    const countries = [
+        ...new Set(results.map(d => d.site.country))
```
samussiah, line 43: Grab the set of countries from `input.dfGroups` directly rather than from `input.dfSummary`.

inst/utils-0.0.1/addCountrySelect.js
```diff
@@ -0,0 +1,77 @@
...
+        const countryOption = document.createElement('option');
+        countryOption.innerHTML = country;
+        countrySelect.appendChild(countryOption);
+        countrySelect.classList.add('country-select');
+    }
+
+
+    // add event listener to dropdown that updates chart
+    countrySelect.addEventListener('change', event => {
+        groupSelect.value = "None";
```
samussiah, line 64: 🥇

inst/utils-0.0.1/addCountrySelect.js
```diff
@@ -0,0 +1,77 @@
...
+        countryFilter =
+        instance.data.config.selectedGroupIDs = results.filter(d => d.site.country === event.target.value).map(d => d.GroupID); // country
+        instance.helpers.updateConfig(instance, instance.data.config, instance.data._thresholds_);
+    });
+
+        // add event listener to dropdown that updates chart
+    groupSelect.addEventListener('change', event => {
+        countrySelect.value = "None";
```
samussiah, line 72: 🥇

inst/utils-0.0.1/addCountrySelect.js
```diff
@@ -0,0 +1,77 @@
...
+    el.style.position = 'relative';
+
+    // add container in which to place dropdown
+    const countrySelectContainer = document.createElement('div');
+    el.appendChild(countrySelectContainer);
+
+    // position container absolutely in upper left corner of `el`.
+    countrySelectContainer.style.position = 'absolute';
+    countrySelectContainer.style.top = 0;
+    countrySelectContainer.style.left = "175px";
```
samussiah, line 24: This looks perfect with three-digit IDs but will overlap with longer IDs.  A `div` that contains both the site select and country select is a more robust approach, something like:

```
<div style = 'position:absolute; top:0'>
    <div class = 'chart-select site-select' style = 'display:inline-block'>...</div>
    <div class = 'chart-select country-select' style = 'display:inline-block'>...</div>
</div>
```

samussiah (changes requested):
This looks really nice @taylorrodgers! I dropped a couple comments below and to summarize it's going to be easiest to add both the site highlight and country highlight with a single function (`addSiteHighlight` say), placing both dropdowns in a single `div`.

It would also be nice to add this control to the bar chart and time series.

taylorrodgers:
> This looks really nice @taylorrodgers! I dropped a couple comments below and to summarize it's going to be easiest to add both the site highlight and country highlight with a single function (`addSiteHighlight` say), placing both dropdowns in a single `div`.
> 
> It would also be nice to add this control to the bar chart and time series.

@samussiah, thank you for looking if over! One question: do we want to add this to the bar and time series chart on this issue / PR? Or would that be a separate ticket? #1543 only mentioned scatterplot. No objection from me if we want to expand the issue to those other charts :)

inst/utils-0.0.1/addCountrySelect.js
taylorrodgers: Hey @samussiah, one potential issue with this that I want to ask about. `input.`dfGroups` does not have a "GroupID" field within it. This isn't an issue with the lines you highlighted above, but that does mean in the example below, I need to hardcode `SiteID` as a mapped value. Is that okay?
image

inst/utils-0.0.1/addCountrySelect.js
taylorrodgers: Hey @samussiah, I made the changes but we'll see the filters overlap with the canvas. In order to work around this problem, we'd have to make changes to `rbm-viz`, I believe. Should we go ahead and log a ticket for that? 

image

samussiah:
> > This looks really nice @taylorrodgers! I dropped a couple comments below and to summarize it's going to be easiest to add both the site highlight and country highlight with a single function (`addSiteHighlight` say), placing both dropdowns in a single `div`.
> > It would also be nice to add this control to the bar chart and time series.
> 
> @samussiah, thank you for looking if over! One question: do we want to add this to the bar and time series chart on this issue / PR? Or would that be a separate ticket? #1543 only mentioned scatterplot. No objection from me if we want to expand the issue to those other charts :)

Yep, give the other charts the same treatment.

inst/utils-0.0.1/addCountrySelect.js
samussiah: This can likely be addressed with inline CSS or even a stylesheet specifically for these dropdowns.

samussiah (approved):
Nice work @taylorrodgers!  Working nicely on my end and you nailed the state maintenance between the two dropdowns.

PR #1629, opened 2024-07-08
Add sample reporting data

### LINKED ISSUE #1626: QC: Create sample reporting data
## QC Details
 create `dfSummary`, `dfGroups`, `dfMetrics` and `dfBounds` sample data that can be used for tests and examples

lauramaxwell:
## Overview
Add sample data for `dfGroups`, `dfMetrics`, `dfSummary` and `dfBounds` for use in reporting workflow examples and tests. Follow the format- `dfGroups` = `sampleGroups`

closes #1626

## Test Notes/Sample Code

Notes:


jonthegeek (commented):
Would it make sense to include some longitudinal data in sampleSummary (different `snapshot_date` values)? I was just sorting out how to create some sample data for #1570, and this would (almost?) entirely cover it if we had some different dates. I guess we'd probably want that reflected in sampleBounds, too? Should this be an entirely separate set of samples (ie, is longitudinal particularly weird)?

lauramaxwell:
@jonthegeek Good call- I think we can make a new `dfSummaryLong` object with a few different snapshot_date values based on what is in `dfSummary` so that it can be used to test longitudinal charts, but we also want to be able to test that those charts get skipped for `dfSummary` objects with only one snapshot date. While we could just filter the long df, to avoid confusion i think i'll just make another small df for longitunidal data. 

Since I'll just be changing dates of the existing data, i don't think i'll need new `dfBounds`, but in the future we plan to add variable thresholds across time (#1595), so depending on that workflow, a longitudinal `dfBounds` might be necessary.

jonthegeek:
@jwildfire pointed out https://github.com/Gilead-BioStats/gsm/blob/dev/inst/examples/LongitudinalReport.R over in #1570 as an option for generating longitudinal data.

jonthegeek (changes requested):
Document these in R/data_specs.R (or elsewhere if that makes more sense).

lauramaxwell:
change sampleMeta back to sampleGroups

lauramaxwell:
@jonthegeek @jwildfire made the edits to the sample data files we discussed yesterday and added docs. also merged in dev, so checks should pass.

jwildfire:
Thanks! Will review and merge now.

jwildfire (changes requested):
New files look good! 

Would be good to do a bit of clean-up in /data and /data-raw. Let's only keep: 
- `gsm::sampleInput`
- `gsm::sampleSummary`
- `gsm::sampleResults`
- `gsm::sampleMetrics`
- `gsm::sampleGroups`
- `gsm::sampleBounds`

Will need to go through and make sure that none of the old data sets are still in use in the package.

Finally, Let's combine all of the R scripts in `\data-raw` into a single `data-raw\sampleData.R`

@jonthegeek do you have time to get started on the clean-up?

jonthegeek:
@jwildfire I should be able to dig into this, and hopefully fix whatever is broken in the GHAs while I'm at it!

jonthegeek:
Datasets that are still used. I didn't delete these/the scripts to generate them. If we want to get rid of them, we should probably explicitly remove their usage outside of this PR (or, alternatively, change the things that use them to use stuff in this PR).

In function defaults:
- meta_workflow (`Widget_SiteOverview()`)

In `test_util-clindata.R`:
- config_param
- config_workflow
- meta_param
- meta_param

In qualification tests:
- config_param

I also created an `example_groups()` function while making `R CMD check` happy. It generates groups for `Widget_SiteOverview()` using `clindata::ctms_site` and `clindata::rawplus_dm`. I strongly suspect that SHOULD be `sampleGroups` or something, but it doesn't match much at all at the moment.

I left `unit_test_data_script.R` for the moment, but IF we want to keep it, it should probably move into the testthat folder (potentially even turned into a helper with functions to load datasets). I strongly suspect we don't need it once we universally use this sampleData, though.

jonthegeek (approved):

jonthegeek:
@jwildfire The column `nMinDenominator` in sampleMetrics isn't mentioned for dfMetrics in the vignette. Should it be? dfMetrics is also missing information in the bulleted list describing it (in both instances). I'm not 100% certain what should be there.

Also see my notes above about datasets that are still used. I suspect they SHOULD be eliminated, but I'm not sure that this is the PR to do that in.

jwildfire:
Thanks John. Agree that we can continue this in another PR. Filed #1642 to deal with removal of legacy data sets and cleaning up dependencies.

jwildfire:
> @jwildfire The column `nMinDenominator` in sampleMetrics isn't mentioned for dfMetrics in the vignette. Should it be? dfMetrics is also missing information in the bulleted list describing it (in both instances). I'm not 100% certain what should be there.
> 
> Also see my notes above about datasets that are still used. I suspect they SHOULD be eliminated, but I'm not sure that this is the PR to do that in.

`nMinDenominator` is a parameter for `Summarize()` so documentation there is probably sufficient. However, the concept of using `workflow$meta` to document custom parameters for a workflow probably does need to be explained in the vignette (we could possibly use `nMinDenominator as a representative example). I'm still planning to do a full read-through of the vignettes soon, so I can keep this on my radar.

jwildfire (approved):

PR #1630, opened 2024-07-09
`Group` > `GroupLevel`

samussiah:
## Test Notes/Sample Code

Notes:


lauramaxwell:
This fixed that `Group` error, but unfortunately another one popped up. Looks like we also need to change `status` to `Status` (and possibly other column name adjustments that we see in the `Params` column in the [wiki](https://github.com/Gilead-BioStats/gsm/wiki/Data-Model#dfmeta))

Issue #1611, opened 2024-06-27
QC: Fix CI/CD pipeline


lauramaxwell:
## QC Details
CI/CD pipeline is currently crashing. find issues and fix them
Known issues: 
- [x] tests are broken due to recent updates (see issues #1620 and #1617)
- [x] #1622
- [x] #1623
- [x] #1624

jonthegeek:
I created #1631 for one of the `R CMD check` notes.

Per Scrum today (2024-07-09), there's no known reason that we care about R 4.0.5 (but the checker currently runs on that). Gilead's Workbench has 4.1.3. Can someone confirm whether any deployments (eg, Connect) are using 4.0.5? @jwildfire @samussiah Note: the tidyverse supports back 4 versions of R, so they won't support 4.0.x once 4.5 comes out in Spring 2025. Ideally anything that's using 4.0.5 should be strongly encouraged to upgrade (and 4.1.3 

If we're aiming for CRAN, we should eventually add `release` and `devel` (current and next versions of R), and probably the ones in-between 4.1 and release while we're at it. We don't need to run all of those checks for every PR, but we can run them when we PR into `main`.

The other note is about objects with no visible binding (ie, bare column names in tidyverse commands). I'll create a separate issue (and associated PR) to properly deal with those.

jonthegeek:
Also see #1633.

samussiah:
I think we're safe to drop `4.0.5`. `4.1.3` is in active use at Gilead though, so we definitely need to keep running it.

PR #1632, opened 2024-07-09
Fix 1631

### LINKED ISSUE #1631: QC: Remove Unused Packages
## QC Details
`R CMD check` flags us for having too many dependencies. This mostly only matters when/if we're aiming to release to CRAN, but, when those packages aren't used, it also makes CI slower, and generally can cause issues for users.

Check each package to determine whether it's actually used:

- [x] arrow
- [x] broom
- [x] cli
- [x] DBI
- [x] dplyr
- [x] ggplot2
- [x] glue
- [x] htmltools
- [x] htmlwidgets
- [x] jsonlite
- [x] kableExtra
- [x] lamW
- [x] lifecycle
- [x] lubridate
- [x] magrittr
- [x] purrr
- [x] rlang
- [x] sqldf
- [x] stats
- [x] stringr
- [x] testthat
- [x] tibble
- [x] tidyr
- [x] tools
- [x] utils
- [x] withr
- [x] yaml

## Additional Comments

We can also check Suggests while we're at it (and maybe move things that are currently in Imports into Suggests):

- [x] clindata
- [x] covr
- [x] desc
- [x] DT
- [x] fontawesome
- [x] gggenes
- [x] ggiraph
- [x] gt
- [x] here
- [x] httr
- [x] knitr
- [x] pander
- [x] pkgdown
- [x] riskmetric
- [x] rmarkdown
- [x] roxygen2
- [x] rstudioapi
- [x] safetyData
- [x] styler
- [x] usethis

### LINKED ISSUE #1633: QC: Split Dev Workflow
## QC Details
`R-CMD-check-dev.yaml` does too much. If we split each thing up into its own workflow, they can run simultaneously for each PR, reducing waiting time. It's also easier to see exactly what fails that way.

## Additional Comments
For the most part, we should be able to literally cut and paste from the yaml.

jonthegeek:
## Overview

Remove unused packages from DESCRIPTION, and move low-use packages to Suggests (closes #1631).

Use data-masking and tidy-select to reference columns in tidyverse functions (toward fixing #1611).

Remove R 4.0.5 from CI.

A few other tweaks to make testing and CI more happy.

Split workflows for easier debugging (closes #1633).


.github/workflows/R-CMD-check-dev.yaml
```diff
@@ -17,7 +17,6 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: ubuntu-latest, r: '4.0.5', repos: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
           - {os: ubuntu-latest, r: '4.1.3', repos: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
```
lauramaxwell, line 20: should we also include latest r release in dev checks since its required in main, to avoid surprises when pushing releases from dev to main?

lauramaxwell (commented):
looks great! one small comment, and we should be gtg

.github/workflows/R-CMD-check-dev.yaml
jonthegeek: Yes, absolutely! I didn't want to change EVERYTHING all at once, but I'll go ahead and do that!

lauramaxwell (approved):
great!

PR #1639, opened 2024-07-11
add example for creating a new missing visit KRI

jwildfire:
## Test Notes/Sample Code

Notes:


jonthegeek (approved):
I reflowed for readability, and moved the workflow to the workflows dir (and removed the absolute path reference). Once it's in the right place, pkgload::load_all causes system.file() to find it.

Overall looks good and makes sense!

jwildfire:
I think this is more of an example than a production workflow. So let's keep it out of inst/workflows for now. I'll put in a quick tweak to get it to load from the examples directory and then merge.

jonthegeek:
@jwildfire This one looks like it's ready to merge!

Issue #1533, opened 2024-04-30
QC: Update Qualification tests for v2


jwildfire:
## Overview

Need to do a full review of Qualification tests for v2. 

## QC Details

Review includes:

- [x] Review/update specs. Consider removing/updating specs if core functionality has changed in v2
- [ ] Review/update list of test cases. Remove test cases that are no longer associated with any tests
- [ ] Draft new test cases. 
- [ ] Review/Update test scripts. I believe that the qualification code will still be valid for many test cases, but the `gsm` code will need to be updated throughout. For example, calls `AE_Assess` will need to be replaced with calls to `Run_Workflow` using a custom workflow object that has the specifications needed for each given test case. 

## Additional Comments

We'll likely want to split this in to multiple issues as we make progress on this.

jonthegeek:
When this is complete, re-enable the pull_request option in `.github/workflows/qualification-tests-dev.yaml` (disabled for #1642).

Issue #1637, opened 2024-07-10
QC: change `dfSummary` in reporting data to `dfResults`


lauramaxwell:
## QC Details
update all reporting workflows, examples and functions that use the reporting version of dfSummary currently

PR #1648, opened 2024-07-11
Remove Legacy Datasets

### LINKED ISSUE #1642: QC: Remove legacy data sets
# Overview

Remove data sets that were supporting gsm v1 workflows and update references to use the new `gsm::sample*` data based on v2 data model. 

# Notes from PR 

Datasets that are still used. I didn't delete these/the scripts to generate them. If we want to get rid of them, we should probably explicitly remove their usage outside of this PR (or, alternatively, change the things that use them to use stuff in this PR).

In function defaults:
- meta_workflow (`Widget_SiteOverview()`)

In `test_util-clindata.R`:
- config_param
- config_workflow
- meta_param
- meta_param

In qualification tests:
- config_param

I also created an `example_groups()` function while making `R CMD check` happy. It generates groups for `Widget_SiteOverview()` using `clindata::ctms_site` and `clindata::rawplus_dm`. I strongly suspect that SHOULD be `sampleGroups` or something, but it doesn't match much at all at the moment.

I left `unit_test_data_script.R` for the moment, but IF we want to keep it, it should probably move into the testthat folder (potentially even turned into a helper with functions to load datasets). I strongly suspect we don't need it once we universally use this sampleData, though.

_Originally posted by @jonthegeek in https://github.com/Gilead-BioStats/gsm/issues/1629#issuecomment-2223327115_

jonthegeek:
## Overview
Deleted old, unused datasets. Also regenerated sampleData to ensure everything was new. Closes #1642.

## Test Notes/Sample Code

Notes: 

I also turned off the `qualification-tests-dev` workflow. I mentioned in #1533 that the workflow should be turned back on once the tests are ready, but let me know if you'd like to do something more complicated to decide when those ones run.


jonthegeek:
@jwildfire Should `sampleSummary` be removed as part of this, per #1637? Right now it looks like `sampleSummary` and `sampleResults` are identical, and I'm not sure if they're supposed to be different.

jwildfire (approved):

jwildfire:
> @jwildfire Should `sampleSummary` be removed as part of this, per #1637? Right now it looks like `sampleSummary` and `sampleResults` are identical, and I'm not sure if they're supposed to be different.

See #1651

PR #1656, opened 2024-07-15
fix #1616

### LINKED ISSUE #1616: Bugfix: Use investigator ID and PI number instead of site ID and site number.
## Expected Behavior
Use investigator ID instead of site ID because the latter isn't necessarily unique in real study data.

## Current Behavior
Referencing `clindata::rawplus_dm$siteid` and `clindata::ctms_site$site_num` instead of `clindata::rawplus_dm$invid` and `clindata::ctms_site$pi_number`.

## Possible Solution
- update column references in workflows and codebase

## Steps to Reproduce

1.
2.
3.
4.

samussiah:
## Overview
- Resolves #1616: replace site ID with investigator ID

## Test Notes/Sample Code

Notes:

### DIFF
--- a/inst/workflow/data_reporting.yaml
+++ b/inst/workflow/data_reporting.yaml
@@ -7,7 +7,7 @@ steps:
     output: dfCTMSSiteWide
     params:
       df: ctms_site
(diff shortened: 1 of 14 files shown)


lauramaxwell (approved):
I made one small unrelated edit to cou0003.yaml, but everything else look good!

Issue #1651, opened 2024-07-13
QC: Update sampleSummary to be a single KRI/Workflow


jwildfire:
`sampleSummary` should be the output of a single KRI workflow and be one record per `GroupID`. In contrast, `sampleResult` should stack results from many KRIs and thus be one record per `GroupID` per `MetricID` per `SnapshotDate`.

_Originally posted by @jonthegeek in https://github.com/Gilead-BioStats/gsm/issues/1648#issuecomment-2225873803_

Issue #1635, opened 2024-07-09
QC: Speed up slow test(s)


jonthegeek:
## QC Details
A handful of test files run considerably slower than the rest. This discourages devs from testing frequently, which can allow bugs to sneak in. Specifically, `util-RunWorkflow.R` took 37.5s on my machine, and `util-clindata.R` took 9.6s. See what might be possible to speed those up.

## Additional Comments
From a very very quick look, I saw there were maps in `test_util-RunWorkflow.R`. It might be possible to subset some of the info and still get the same test coverage. Also check whether it makes sense to pre-build any of the artifacts that aren't directly involved in that test (and save them as RDS).

PR #1657, opened 2024-07-15
Standardize gt usage

### LINKED ISSUE #1652: QC: Use `util-gt.R` in other report tables
## QC Details
Let's standardize usage of `util-gt.R` by updating `Report_MetricTable()` and maybe `Report_StudyInfo()` to use the same approach as `Report_FlagOverTime()`.

jonthegeek:
## Overview
Standardize usage of {gt}.
Closes #1652.

## Test Notes/Sample Code

Notes: 

As a step 0, I fixed the dates in sampleSummary, and reconciled the tests with the updates from previous PRs (changing to the sample datasets broke a bunch of things). I wanted tests to pass before I make changes, so I can tell for sure that I didn't break anything.


jonthegeek:
@jwildfire @lauramaxwell I fixed the broken tests in this to clear the field before I move on to the gt stuff. Can one of you confirm that my changes make sense? A number of tests were looking for specific values, but I'm PRETTY sure those values legitimately changed with the new data.

There's one actual bugfix in there, in `Report_Setup.R`. It was taking the first date and THEN finding max of that, and I think you meant for it to be the max of the whole column.

Of course now I see a test failing on here so I'll check that before moving on. They SHOULD all pass...

jonthegeek:
Update: Tests weren't broken, but a number of other things were throwing errors in R CMD check. This version should work. Once the checks are green, I'll implement my feature and make sure I don't break anything.

jonthegeek:
Issues #1659 and #1660 deal with more complex kable-to-gt cases. This PR does a fair amount of cleaning/refactoring, but the main thing it implements is a standardization of gt() calls (in `Report_StudyInfo()`).

data-raw/sampleData.R
```diff
@@ -24,5 +24,7 @@ rm(sampleResults)
 
 sampleSummary <- read.csv("data-raw/sampleSummary.csv")
 sampleSummary$GroupID <- as.character(sampleSummary$GroupID)
+sampleSummary$SnapshotDate <- as.Date(sampleSummary$SnapshotDate,
```
lauramaxwell, line 27: i don't think we want sampleSummary to have a snapshot date- this is the output of `Summarize()` before it gets transformed for reporting. #1661 updates the sampleData file per #1651

tests/testthat/test_util-RunWorkflow.R
```diff
@@ -20,11 +20,18 @@ lData <- gsm::UseClindata(
   )
 )
 
+# Quiet output
+quiet_RunWorkflow <- function(...) {
```
lauramaxwell, line 24: we used to have a `bQuiet` parameter in these functions. should we go that route again, or is this wrapper function preferred? @jwildfire

lauramaxwell (changes requested):
overall- great! just a few small details we can discuss. see the comments

data-raw/sampleData.R
jonthegeek: Yeah, this was to get things working-enough before #1661. I think we should get that and the test-fixing PR merged and then I can merge them into here and resolve conflicts.

tests/testthat/test_util-RunWorkflow.R
jonthegeek: This was definitely aiming at "good enough for now." I needed to quiet down this test so I could more easily see what was actually failing. I *really* want to work on revising this set of tests to speed things up (#1635), and doing so in the context of any desired refactors (like perhaps adding this argument) would definitely make sense.

data-raw/sampleData.R
jonthegeek: At this point in the lifecycle, this line of code is needed to get tests to pass. Can we merge this PR into dev, and then merge dev into #1661 and reconcile? I can do the conflict-resolution, I'd just really like to get all checks back to green and keep them there going forward!

tests/testthat/test_util-RunWorkflow.R
jonthegeek: Thinking it through... I know I find it annoying that {httr2} has a message about using the cache that can't (easily) be silenced, and thus fills up a log for something I run with it. I'd definitely appreciate being able to change the verbosity level. 

Per [Hadley's in-progress book](https://design.tidyverse.org/boolean-strategies.html) and [some discussions I've seen](https://fosstodon.org/@coolbutuseless/112742297912462306), it might be better to use  enumerated values rather than a boolean, in case we want to have multiple levels in the future. Something like `strVerbosity = c("verbose", "quiet")`

jwildfire: @lauramaxwell  @jonthegeek As much as I loved having something called `bQuiet`, I definitely don't want to go down that route again (it was a fairly major pain to maintain). I'm fine with using `suppressMessages()` in testing as needed.

lauramaxwell (approved):
looks good!

Issue #1628, opened 2024-07-08
QC: Update headers


lauramaxwell:
## QC Details
Update all headers for all functions in `gsm`.
- Ensure that examples work properly and are as simple as possible (ex. use sample data instead of running through the full workflow to create charts and reports). 
- Maintain consistency between help files for content and structure

jonthegeek:
See #1658 for a way to maintain consistency for parameters, at least.

PR #1667, opened 2024-07-16
Implement a glossary

### LINKED ISSUE #1658: QC: Shared parameter documentation
## QC Details
For something like this, with a number of parameters that are reused across many functions, I find it helpful to document those parameters in a central place. For example, it can be helpful to have a `aaa-shared_params.R` file containing a named `shared-params` object with `@param` definitions, and then all functions that reference those shared parameters use `@importFrom shared-params` to get the central, canonical definition of that parameter.

jonthegeek:
## Overview
Glossary added as `man/glossary`. Definitions are reusable for things such as function parameters or vignettes.

Closes #1658.

## Test Notes/Sample Code

Notes: 

@jwildfire I played with this a little, and have it *mostly* solved for a reusable glossary. Info from [R Packages](https://r-pkgs.org/man.html#child-documents) and [roxygen2 help](https://roxygen2.r-lib.org/articles/reuse.html#child-documents). However, I tried specifying the link like they show there, and it *always* applies (not selectively like the help implies), thus the links to other functions were broken. I'll dig into it a bit more + see if they have advice tomorrow. I suspect it's a bug, possibly specifically in the context of `@param`.


R/Report_FlagOverTime.R
```diff
@@ -3,9 +3,8 @@
 #' Create a table of longitudinal study data by site, study, or country, showing
 #' flags over time.
 #'
+#' @inheritParams shared-params
```
lauramaxwell, line 6: This is great! so the idea is that _all_ params in the `shared-params` header would be popped into the given function's header? in which case, we'd probably want a series of `shared-params` objects for different common combinations of params. or is it possible to selectively include them by specifying the name of the param?

jonthegeek: When you @inheritParams, roxygen automatically only grabs the params that are used in the function. So that one object can include *all* params, and roxygen will just grab the ones that apply.

R/aaa-shared.R
```diff
@@ -0,0 +1,12 @@
+#' Parameters used in multiple functions
+#'
+#' Reused parameter definitions are gathered here for easier usage. Edit the
+#' definitions in `man/glossary/{term}.Rmd` (one file per term).
+#'
+#' @param dfMetrics
+#' ```{r, child = "man/glossary/dfMetrics.Rmd"}
```
lauramaxwell, line 7: alternatively, we could just include this type of snippet in headers that have non-standard param configurations, so the description is still centralized.

R/Report_FlagOverTime.R
lauramaxwell: oh nice!

jonthegeek:
Sadly it does appear to be a bug, which I've [reported to the roxygen2 team](https://github.com/r-lib/roxygen2/issues/1637). I'm going to move forward with the idea anyway, and assume that we'll be able to fix the handful of files one way or another. For now I'll write them in the manner that will work for documentation, and then we can later make them ALSO work for vignettes when/if they address my issue.

jonthegeek:
@jwildfire @lauramaxwell I think I like this general idea a lot! Things to check before I push forward:

- The general process for any new parameter (or at least once that parameter will be used a second time):
  - Create a new YAML file in `man/glossary`, such as `man/glossary/dfMetrics.yaml`
  - Add a `class` and `definition` within that YAML. The `definition` should describe the parameter (no need to describe the class here) and how to produce it. See `man/glossary/dfMetrics.yaml` and `man/glossary/dfResults.yaml` (or any of the files where their definitions are used) for examples.
  - Add that parameter to `R/aaa-shared.R` with a new line: ``@param {paramName} `r param_from_yaml("{paramName}")` `` (eg, ``@param dfMetrics `r param_from_yaml("{dfMetrics}")` ``
  - Add `@inheritParams shared-params` to the top of the `@param` block in your function (and anything else that uses that param). Eventually every function will presumably include this line. Note: If you don't want to use the "default" definition of a parameter in a given function, you can include a full `@param` definition for that parameter, and it will override the inherited result. `@inheritParams` also only inherits params that are relevant to the particular function (it won't add documentation for EVERYTHING that's in the `shared-params` block).
  - Once we nail this down, it might make sense to split out functions to create these yamls. I can definitely imagine this being a simple package in its own right (for creating and using the definitions).
- It probably wouldn't be a bad idea to pull this PR down locally, `devtools::load_all()`, and `?Report_KRI` to see how the definitions of `dfResults` and `dfMetrics` look in context.

Note that, if we decide to change the definitions, we only have to change them in one place (so no big deal)... but I plan to do this for all of our reused parameters, so we should probably figure out the general idea to minimize re-working.

What do you think?

jonthegeek:
I think this is actually "done enough" for #1658. Applying the glossary will be at least one push as part of #1628, but this specific ticket shouldn't hold off until every parameter is re-documented. Let's get this one sorted out as its own PR, and THEN we can decide how to systematically apply it.

lauramaxwell (commented):
I think this will be helpful for centralizing the documentation for df inputs and a few other unambiguous params, like `vThreshold``. One thing i'm finding through reviewing this is that we use the same param name for things across different functions that aren't identical, such as `strGroupCol` and `strType`. Not sure that this a major issue, just wanted to bring it up before we go too far down this path. It seems the use for the glossary will be limited to under ten params, but centralization will still ultimately probably be super helpful.

jonthegeek:
> I think this will be helpful for centralizing the documentation for df inputs and a few other unambiguous params, like `` vThreshold``. One thing i'm finding through reviewing this is that we use the same param name for things across different functions that aren't identical, such as ``strGroupCol`and`strType`. Not sure that this a major issue, just wanted to bring it up before we go too far down this path. It seems the use for the glossary will be limited to under ten params, but centralization will still ultimately probably be super helpful.

Check out discussion #1682 for something related to this. You're definitely right, though; sometimes it makes sense to use the same name for at least slightly different meanings.

I have one more idea to experiment with for cases where there's a shared core, but possibly slightly different specifics. {roxygen2} doesn't allow you to @importParams and enhance, but I have an idea for how to make that work (so, for example, in the `dfTransformed` we could use a shared definition that has all the generic information, and then tack on a bit about which function we expect you to use to make it for this function). 

I don't want to force unnatural rigidity or anything, but it's so easy for definitions to drift apart if we aren't vigilant, and that can be really confusing as a user. Plus it's tedious to have to describe the same basics about a parameter that we've described before.

lauramaxwell (approved):
I think this will work for us. I'm going to merge this in and work on the update headers issue with the intention of implementing this strategy as much as possible.

PR #1679, opened 2024-07-18
Update examples

### LINKED ISSUE #1655: QC: Clean up `inst/examples`
## QC Details

- Use new `gsm::sampleXXX` data to simplify examples when appropriate
- add examples for running country level workflows
- add example for running multiple snapshots
- QC all examples to make sure they run
- sync with cookbook vignette

jwildfire:
## Overview

Refactor examples. 

closes #1655 

## Test Notes/Sample Code

Notes:


lauramaxwell:
I like this format for the examples. I think i'll work on editing the examples a bit to add some `spin` markup tags so that we can use `knitr::spin_child()` to include the examples in `Cookbook.Rmd`. This will allow us to  break up the examples within each example file with markdown text instead of just R comments while keeping these example files as just `.R` scripts instead of `.Rmd`s.   Will probably be a v2.1.0 issue, but i think it will make the cookbook look a little nicer

lauramaxwell (approved):
had to update a test snapshot, but these all look good and are clear to me!

Issue #1680, opened 2024-07-18
QC: Test workflows


jonthegeek:
## QC Details
Run all workflows to make sure we haven't broken them. Maybe as a test, or maybe as a separate GHA.

PR #1683, opened 2024-07-18
Faster RunWorkflow() tests

### LINKED ISSUE #1635: QC: Speed up slow test(s)

jonthegeek:
## Overview
`test-util-RunWorkflow.R` now takes about 10s on my system, vs ~40s previously.

Closes #1635.

## Test Notes/Sample Code

Notes: 

We should separately test all workflows per #1680, but we don't need to run those tests every time we change things. The general functionality tests SHOULD run often, though, so running ~fast is helpful.


lauramaxwell (approved):
nice! looks good to me.

PR #1702, opened 2024-07-23
Update headers with glossary and simplified examples

### LINKED ISSUE #1628: QC: Update headers

lauramaxwell:
## Overview
closes #1628

## Test Notes/Sample Code

Notes:


man/glossary/dfBounds.yaml
```diff
@@ -0,0 +1,8 @@
+class: data.frame
+description: >
```
jonthegeek, line 2:
```suggestion
definition: >
```

man/glossary/dfBounds.yaml
```diff
@@ -0,0 +1,8 @@
...
+  Set of predicted percentages/rates and upper- and
+  lower-bounds across the full range of sample sizes/total
+  exposure values for reporting. Created by passing
+  `dfResults` and `dfMetrics` to `MakeBounds()`. Expected
```
jonthegeek, line 6:
```suggestion
  `dfResults` and `dfMetrics` to [MakeBounds()]. Expected
```

man/glossary/dfGroups.yaml
```diff
@@ -0,0 +1,5 @@
+class: data.frame
+description: >
```
jonthegeek, line 2:
```suggestion
definition: >
```

man/glossary/dfInput.yaml
```diff
@@ -0,0 +1,6 @@
+class: data.frame
+description: >
```
jonthegeek, line 2:
```suggestion
definition: >
```

R/util-MakeBounds.R
```diff
@@ -16,9 +17,12 @@
 #' @return A data frame.
 #'
 #' @examples
+#'  \dontrun{
```
jonthegeek, line 20: Why `/dontrun`? It feels like this should run fine with the sample data.

jonthegeek (changes requested):
Typo'ed name in glossary files.

For all the `\dontrun`'s, consider switching to `@examplesIf interactive()` (so the example doesn't run in checks but is easy for a user to play with), or eliminate the `\dontrun` entirely in some cases, I think.

R/util-MakeBounds.R
lauramaxwell: yeah, i'll delete a bunch of these `dontrun`s. not sure what i was thinking this morning when i added all of them...

R/Report_KRI.R
```diff
@@ -1,77 +1,32 @@
...
 #' strOutpath <- "StandardSiteReport.html"
 #' Report_KRI(
 #'   lCharts = lCharts,
-#'   dfSummary = dfSummary,
-#'   dfSite = dfSite,
-#'   dfStudy = dfStudy,
-#'   dfMetrics =dfMetrics,
+#'   dfResults =  reportingResults,
+#'   dfGroups =  reportingGroups,
+#'   dfMetrics = reportingMetrics
```
jonthegeek, line 27:
```suggestion
#'   dfMetrics = reportingMetrics,
```
This comma is throwing the styling error, I think.

jonthegeek (changes requested):
I'm pretty sure this missing comma is what's making the styler sad. Re-style after this fix and see.

lauramaxwell:
@jonthegeek This is ready for re-review. Along with a few typos and `\dontrun`s being removed, I also had to fix most of the visualization and widget examples to used filtered versions of the `reporting` data frames so that only one metric and snapshot date (if applicable) was being fed into the functions. This wasn't caught earlier because it doesn't throw an error- it just creates a chaotic plot. Should we add a safeguard against this in the form of an error message? or are we assuming that most people are using `MakeCharts()` and so they don't have to worry about this filtering typically? @jwildfire @samussiah

lauramaxwell:
we could also auto-filter to latest date and first metric in these functions and provide an alert/message that lets the user know what happened. This way, the output will always be reasonable, and if the user wanted something different, they will be aware of the filtering that occured.

jonthegeek:
@lauramaxwell We ran into this issue in the POC I'm updating for Friday's meeting, in `Widget_GroupOverview()`. There we're manually filtering to `dplyr::filter(SnapshotDate == max(SnapshotDate))`, but it would probably be good to build that sort of check into everything. I know at least one function does it right now. I think @jwildfire was thinking we'd standardize that for 2.1.

jonthegeek (approved):
Looks good! There was one remaining issue in Report_KRI.R (an extraneous space before `\dontrun`) that was causing styler to complain, but I was able to clear that up. I had to re-document, which is another point for checking documentation in our standard checks (#1694), and/or just auto-documenting PRs.

PR #1731, opened 2024-07-29
Fix overall group select and group overview sorting.

### LINKED ISSUE #1723: Bugfix: Interaction between overall group select and site/country select can error out.
## Expected Behavior
Overall group select syncs up widget-level group selects and sets all country selects to `"None"`.

## Current Behavior
image

## Possible Solution
distinguish between site and country select in the overall group select logic:

```
    if (event.target.value !== 'None') {
      document.querySelectorAll(".gsm-widget-control--select").forEach((el) => {
        el.options[el.selectedIndex].innerHTML = event.target.value;
        el.disabled = true;
      });
    } else {
      document.querySelectorAll(".gsm-widget-control--select").forEach((el) => {
        el.options[el.selectedIndex].innerHTML = "None";
        el.disabled = false;
      });
    }
```

## Steps to Reproduce

1.
2.
3.
4.

### LINKED ISSUE #1730: Bugfix: sorting in the overview table widget
## Expected Behavior
sorting in the metric columns of the overview table widget should sort based on flag severity when clicked

## Current Behavior
sorts on `abs(score)` which isn't consistent with flags because negative values typically have a lower threshold

## Possible Solution
change sort to `flag`  or `flag` and `score` for the metric columns

## Steps to Reproduce

1.
2.
3.
4.

samussiah:
## Overview
- Resolves #1730 
- Resolves #1723 

## Test Notes/Sample Code
- In the group overview table metric columns should sort red > amber > green on the first click and the reverse on the second click.
- The overall group select should highlight the selected group ID in all charts.
- Group and country selects in each chart should not be disabled, i.e. they should work independently of the overall group select.
- When the overall group select is set to _None_ all chart selects should reset to _None_, including country selects.

Notes:


jonthegeek (approved):
This makes sense & works. I'll leave it unmerged until I know what @lauramaxwell wants, but it's good to go when she wants it!

lauramaxwell (approved):
looks good and works well for me. thanks!

PR #1741, opened 2024-07-30
Simplify `R-CMD-check-dev.yaml` matrix

### LINKED ISSUE #1737: QC: Simpler check matrix
## QC Details
As a dev, I'd like my PR tests to complete faster, so that I'm not spending too much time waiting.

We don't need thing to run all the time on every browser. It's ok for us to find out about browser weirdness after we merge a PR; we can fix things before release.

jonthegeek:
## Overview
- Simplify R-CMD-check-dev matrix.
- Remove unused (I think, checks will verify) ssh-agent step from all GHA workflows.

## Test Notes/Sample Code

Notes: 

Closes #1737

### DIFF
--- a/.github/workflows/R-CMD-check-dev.yaml
+++ b/.github/workflows/R-CMD-check-dev.yaml
@@ -17,8 +17,6 @@ jobs:
       fail-fast: false
       matrix:
         config:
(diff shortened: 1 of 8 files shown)


lauramaxwell (approved):
makes sense to me! thanks.

PR #1746, opened 2024-07-31
Prototype Data Specification Framework

jwildfire:
## Overview

WIP. See #1744 for overview.

## Test Notes/Sample Code

Run code in `inst/examples/3_ReportingExamples.R` to test.

Notes:


jwildfire:
@lauramaxwell @samussiah @dpastoor  this is ready for a first look.

R/util-checkSpec.R
```diff
@@ -0,0 +1,66 @@
+# Check that all data.frames and columns in the spec are present in the data
+#' Function to check if the data and spec are compatible
+#'
+#' This function checks if the data and spec are compatible by comparing the data.frames
+#' and columns in the spec with the data.
+#'
+#' @param lData A list of data.frames.
+#' @param lSpec A list s specifying the expected structure of the data.
```
lauramaxwell, line 8: Would it be useful to have column data type in this `lSpec` list as well? This would allow us to confirm not just structure but also ensure that the content is what we expect. could be an optional parameter. So, when used, it would look something like:
```
Raw_QUERY:
  subjectname: 
     required: true
     datatype: character
 querystatus: 
     required: true
     datatype: character
  queryage:
     required: true
     datatype: integer
```

samussiah: I like this better as well, and the current implication only considers the names anyway so it works.

R/uti-CombineSpecs.R
jonthegeek, line 1: Filename should be `util-CombineSpecs.R` (missing `l` right now) to match other utils.

R/uti-CombineSpecs.R
```diff
@@ -0,0 +1,25 @@
+#' Combine Specifications
+#'
+#' This function combines a list of specifications into a single specification.
```
jonthegeek, line 3:
```suggestion
#' Combine a list of specifications into a single specification.
```
No need to include "this function" in the description.

R/uti-CombineSpecs.R
```diff
@@ -0,0 +1,25 @@
...
+#'
+#' @param lSpecs A list of specifications.
+#' @return A list representing the combined specification.
+#' @examples
+#' all_wf <- gsm::MakeWorkflowList()
```
jonthegeek, line 8:
```suggestion
#' all_wf <- MakeWorkflowList()
```
Don't need the namespace within the package (and having it confuses RStudio if you try to F2 from function definition to function definition).

R/uti-CombineSpecs.R
```diff
@@ -0,0 +1,25 @@
...
+
+CombineSpecs <- function(lSpecs){
+    all_specs <- list()
+    for (spec in lSpecs){
+        for (domain in names(spec)){
+            if (domain %in% names(all_specs)){
+                all_specs[[domain]] <- c(all_specs[[domain]], spec[[domain]])
+            } else {
+                all_specs[[domain]] <- spec[[domain]]
+            }
```
jonthegeek, line 21:
```suggestion
            all_specs[[domain]] <- c(all_specs[[domain]], spec[[domain]])
```
The if/else is redundant. `all_specs[["new_domain"]]` is `NULL`, and `c(NULL, spec[["new_domain"]]` is `spec[["new_domain"]]`, so they can combine this way in either case.

R/uti-CombineSpecs.R
```diff
@@ -0,0 +1,25 @@
...
+        }
+    }
+    return(lSpec)
```
jonthegeek, line 24:
```suggestion
    return(all_specs)
```

R/util-checkSpec.R
```diff
@@ -0,0 +1,66 @@
+# Check that all data.frames and columns in the spec are present in the data
+#' Function to check if the data and spec are compatible
+#'
+#' This function checks if the data and spec are compatible by comparing the data.frames
+#' and columns in the spec with the data.
```
jonthegeek, line 5:
```suggestion
#' Check if the data and spec are compatible
#'
#' Check if the data and spec are compatible by comparing the data.frames and
#' columns in the spec with the data.
```

R/util-checkSpec.R
```diff
@@ -0,0 +1,66 @@
...
+        lDataColumns <- names(lData[[strDataFrame]])
+        allCols <- c(allCols, paste(strDataFrame, lSpecColumns, sep="$"))
+
+        thisMissingCols <- lSpecColumns[!lSpecColumns %in% lDataColumns]
+        if(length(thisMissingCols) > 0){
+            missingCols <- c(missingCols, paste(strDataFrame, thisMissingCols, sep="$"))  
+        }   
+    }
+    if (length(missingCols) > 0) {
+        cli_alert_danger("Not all columns in the spec are present in the data, missing columns are: {missingCols}")
```
jonthegeek, line 61:
```suggestion
        cli_abort("Not all columns in the spec are present in the data, missing columns are: {missingCols}")
```
See above.

R/util-checkSpec.R
```diff
@@ -0,0 +1,66 @@
...
+#'
+#' @export
+#' 
+CheckSpec <- function(lData, lSpec) {
+    # Check that all data.frames in the spec are present in the data
+    lSpecDataFrames <- names(lSpec)
+    lDataFrames <- names(lData)
+    if (!all(lSpecDataFrames %in% lDataFrames)) {
+        MissingSpecDataFrames <- lSpecDataFrames[!lSpecDataFrames %in% lDataFrames]
+        cli_alert_danger("Not all data.frames in the spec are present in the data, missing data.frames are: {MissingSpecDataFrames}")
```
jonthegeek, line 42:
```suggestion
        cli::cli_abort(c(
            "{.arg lData} must contain all data.frames in {.arg lSpec}.",
            i = "Missing data.frames: {MissingSpecDataFrames}"
        ))
```
`cli_alert_danger()` doesn't actually error, it just tells the user that something bad happened. If we're checking a spec, we should stop with an actual error.

I prefer [tidyverse style for error messages](https://style.tidyverse.org/error-messages.html), which is basically a short, clear problem statement on one line, and then details. I included that format here, feel free to ignore that part of the suggestion if you want to stick with the one-liner.

jonthegeek (changes requested):
I really like this plan in general!

I agree that it would be nice to allow for optional class specification. We should coordinate that with #1750, and use the same checking system.

R/uti-CombineSpecs.R
samussiah: Copilot feels otherwise - "the" is its favorite word.

R/util-checkSpec.R
jwildfire: Yep, I'm ok with the slightly more complex implementation. Will update.

jwildfire:
Made the suggested changes, if someone can add unit tests for `CombineSpecs` and `CheckSpec, I think we can merge. 

Let's file separate issues for: 
- Update CheckSpec to take required parameter into account
- Update specs to add type parameter and incorporate in CheckSpec

And cover in a new PR. 

@lauramaxwell can you coordinate please.

R/util-CombineSpecs.R
```diff
@@ -0,0 +1,21 @@
...
+#' all_wf <- MakeWorkflowList()
+#' all_specs <- CombineSpecs(all_wf)
+#'
+#' @export
+
+CombineSpecs <- function(lSpecs){
+    all_specs <- list()
+    for (spec in lSpecs){
+        for (domain in names(spec)){
+            all_specs[[domain]] <- c(all_specs[[domain]], spec[[domain]])
```
lauramaxwell, line 17: When writing tests for this, I thought of a scenario that i want to make sure i understand the intended use. If `col1` is in both domain 1 and domain 2, then should we (1) de-dupe this list, (2) throw a message in case they are named the same but aren't the same, or (3) just keep all duplicates? Currently the third option is how it behaves.

jwildfire: I think we just de-dupe the list. 

More generally, I think it's generally fine (even expected) to have duplicate columns across domains/specs. `siteid` and `subjectid` for example, would be expected in multiple domains. Similarly, it would be fine for a column to be included in specs for the same domain in multiple modules For example, {gsm.app} and {gsm.endpoints} would both expect dfAE to have a `subjectid` column.

Is there another use case you have in mind that I'm not considering?

jwildfire: I guess there is a bit of nuance we might need around a scenario where the underlying properties of a column are different across 2 different specs. I guess we would want to do the following: 
- if one spec has `required: true` and the other has `required: false` set the combined spec to `required:true`
- if specs have different `type` the we should use the first `type` and throw a warning about the mismatch. 

I think these probably belong in a separate v2.1 issue.

lauramaxwell: yes, i think we should de-dupe. and as you mention, once we have more properties to a column, we would want to make sure they match (in the case of `type`). for `required` it might be slightly more complicated because we wouldn't want to error out when an optional column is missing from a given df even if it is required elsewhere. i'll file a few issues to address these edits.

lauramaxwell:
vignette builds and examples are failing because of new `lSpec` requirement. will update this in the morning so we merge this in with all checks passing.

lauramaxwell (approved):
Checks are all passing now after some updates to vignettes, examples and tests to incorporate new df naming conventions. Issues  #1777 #1778 and #1779 have been added to address discussion about `CombineSpec()` improvements and wider implementation across metric yamls.

Overall, great start!

PR #1751, opened 2024-08-06
Fix workflow typo.

### LINKED ISSUE #1734: QC: Only run pkgdown-dev on push *to* dev
## QC Details
The PR of `dev` into `main` (#1732) triggered `pkgdown-dev.yaml` to run. That is intended to run when something is pulled *into* dev, not for any PR involving `dev`. Update the action to only run when a PR is merged into `dev`.

jonthegeek:
## Overview
Despite what I said in #1734, this doesn't appear to be triggered specifically because of the dev-to-main PR; it just looked like it because that PR "changes" whenever something merges into dev. What's really happening here is that there was a typo, which was causing this workflow to never actually (successfully) run (except once or twice early on when I was testing things).

Closes #1734.

## Test Notes/Sample Code

Notes:

### DIFF
--- a/.github/workflows/pkgdown-dev.yaml
+++ b/.github/workflows/pkgdown-dev.yaml
@@ -1,6 +1,6 @@
 on:
-   push:
-     branches: dev
+  push:
+    branches: dev
   workflow_dispatch:
 
 name: pkgdown-dev


lauramaxwell (approved):
👍

PR #1732, opened 2024-07-29
gsm v2.0.0-RC-1 - Code Review

jwildfire:
## Overview

This is a PR for reviewing updates in gsm v2 that won't be merged. We'll create a separate PR for the final release.

PR #1754, opened 2024-08-06
Build sample reports.

### LINKED ISSUE #1640: QC: Sample Reports as Articles
## QC Details
Instead of building the sample reports with special tricks in the `pkgdown.yaml` workflow, add them as articles. The `_pkgdown.yml` will need to be updated to put these articles into the correct place, but it should be possible to streamline things that way.

jonthegeek:
## Overview
Add a GHA to build the FULL site on PR to main, so we can make sure sample reports work.

Closes #1640

## Test Notes/Sample Code

Notes: 

Do NOT merge this. I'm temporarily building sample reports on every PR to dev with this, but I think at most we want to do this on PR to main, merge to main, and MAYBE merge to dev.

### DIFF
--- a/.github/workflows/pkgdown-dev2main.yaml
+++ b/.github/workflows/pkgdown-dev2main.yaml
@@ -34,6 +34,7 @@ jobs:
         shell: Rscript {0}
         run: |
           devtools::load_all()
+          dir.create(here::here("pkgdown", "assets"), recursive = TRUE)
 
           lData <- list(
             dfSUBJ = clindata::rawplus_dm,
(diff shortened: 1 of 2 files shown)


jonthegeek:
@lauramaxwell This is all set. Unfortunately you can't REALLY test it (since the new one only runs on PR to main), but it might run *after* you merge it (since we have a PR into main over in #1732).

After we release, we may need one more update to properly tag the prod site as "release". See https://github.com/Gilead-BioStats/gsmApp/issues/147 for some notes about that. I don't think we have any release-specific tagging in this one so it shouldn't matter, but I'll continue to look into what might be missing.

PR #1776, opened 2024-08-12
Release v2.0.1

lauramaxwell:
## Overview
Minor patch release for reporting issues noted in the draft release notes [here](https://github.com/Gilead-BioStats/gsm/releases/edit/untagged-4b70ca7a89227dee28ca)

## Test Notes/Sample Code

Notes:


R/util-RenderRmd.R
```diff
@@ -0,0 +1,32 @@
...
+  # specify strOutputDir path, depending on write access to strOutputDir
+  if (file.access(strOutputDir, mode = 2) == -1) {
+    tpath <- fs::path_temp()
+    cli::cli_inform("You do not have permission to write to {strOutputDir}. Report will be saved to {tpath}")
+    strOutputDir <- tpath
+  }
+  rmarkdown::render(
+    input = strInputPath,
+    output_file = file.path(strOutputDir, strOutputFile),
+    intermediates_dir = fs::path_temp(),
```
dpastoor, line 28:
```suggestion
    # this intermediates dir is required in situations where the gsm library is
    # installed in a read-only directory, as the intermediate knit.md file could then
    # not be written when calling rmarkdown::render on any system.file'd rmarkdown taht existed in the package.
    # this makes sure that intermediate content is not written if the original
    # rmd is in such a potentially read-only location
    intermediates_dir = fs::path_temp(),
```

dpastoor: figured I'd type up this code comment for some historical context in case people later on revisit the helper and wonder why we're doing this.

jwildfire (approved):
Looks great

Issue #1781, opened 2024-08-15
Feature: Separate Each Report in to a separate Workflow YAML


jwildfire:
## Feature Details

Each report should be listed separately in `inst/workflow/report` and `snapshot.yaml` should be updated to call reports. 

See #1771 for background.

Issue #1678, opened 2024-07-18
QC: Eliminate use of cli_alert_*()


jonthegeek:
## QC Details
I finally found [confirmation that `cli_alert_*()` functions are leftovers from an older way of doing things in {cli}](https://github.com/r-lib/devtools/issues/2430). Per that ticket:

- [ ] "cli_alert_danger" = "use cli::cli_inform()" (but I'd guess we probably actually mean `cli_abort()` or `cli_warn()`)
- [ ] "cli_alert_info" = "use cli::cli_inform()"
- [ ] "cli_alert_success" = "use cli::cli_inform()"
- [ ] "cli_alert_warning" = "use cli::cli_inform()" (but I assume we actually mean `cli_warn()`)

The main issue I've found is that `cli_alert_warning()` doesn't actually count as a warning for testing; it's just a message dressed up as a warning. Not a HUGE issue but can make things confusing.

jwildfire:
Makes sense to me to tighten this up. Let's address in v2.1 or v2.2. We may also want to do a more through review of our logging as we move more runs in to GHA.

lauramaxwell:
Based on discussion in #1814 , we should replace `cli_alert_danger()` with `cli_abort()` in most cases so that we don't continue in a pipeline that has failed validation-type checks.

lauramaxwell:
duplicate of #1933

PR #1783, opened 2024-08-15
Reporting Workflow Refactor

### LINKED ISSUE #1780: Feature: Add support for `context` in workflows
## Feature Details
Optionally Allow a user to specify a list where parameters can be found. This location should take priority over the standard hierarchy (workflow$meta, workflow$data, etc)

jwildfire:
## Overview

Separate reports into separate workflows and update example. See #1771  for additional context. 

Fix #1780 #1781

## Test Notes/Sample Code

Run updated example 3.2

Notes:


lauramaxwell:
I'm going to hold off on fixing all the tests/checks until we have `mapping_refactor` branch merged in

R/RunStep.R
```diff
@@ -42,18 +42,29 @@ RunStep <- function(lStep, lData, lMeta) {
 
   cli::cli_h3("Evaluating {length(params)} parameter(s) for {.fn {lStep$name}}")
 
-  # This loop iterates over each parameter in the 'params' object.
+  # Allow setting parameters from a list in lData when the 'context' parameter is present in the 'lStep' object.
+  if(!is.null(lStep$context)) {
```
lauramaxwell, line 46: @jwildfire Do we have an example of how this `context` field would work? it's in both `lStep` and `lData`, but it seems that the bulk of the information would be in the `lData` element. I'm just not sure i'm completely following how `context` would be used-- seeing an example may be helpful.

inst/workflow/metrics/cou0002.yaml
```diff
@@ -15,7 +15,7 @@ spec:
   Mapped_SeriousAE:
```
lauramaxwell, line 15: Should this be `Mapped_AE`, since we are doing the filtering in `Input_Rate()` now?

inst/workflow/data_mapping.yaml
```diff
@@ -1,141 +1,54 @@
...
-      df: Raw_ENROLL
-      strQuery: "SELECT * FROM df"
+      strQuery: "SELECT subjectname as subject_nsv, data_entry_lag FROM df"
   - name: RunQuery
-    output: Mapped_ScreenFail
+    output: Mapped_SCREEN
     params:
       df: Raw_ENROLL
-      strQuery: "SELECT * FROM df WHERE enrollyn = 'N'"
+      strQuery: "SELECT subjid, invid, country, enrollyn FROM df"
```
jonthegeek, line 54:
```suggestion
      strQuery: "SELECT subjid, invid, country, enrollyn FROM df"

```
Include a newline at the end of the file. Otherwise `readLines()` throws a warning. RStudio has a setting under Code > Saving, "Ensure that source files end with newline", pretty much entirely to avoid triggering that warning.

R/RunStep.R
```diff
@@ -42,18 +42,29 @@ RunStep <- function(lStep, lData, lMeta) {
...
+    if(is.list(lData$context)){
+      context <- lData$context
+    }else{
+      context <- NULL
+    }
+  }
```
jonthegeek, line 52:
```suggestion
```
We don't need this unless we trigger the if conditions below.

jonthegeek (changes requested):
A few changes. Still need to dig through tests.

lauramaxwell:
@jonthegeek i'm working on getting tests to pass

tests/testthat/test-util-checkSpec.R
```diff
@@ -58,3 +58,35 @@ test_that("Multiple missing columns are correctly reported", {
...
+  lSpec <- list(
+    reporting_groups = list(
+      GroupID = list(required = TRUE),
+      GroupLevel = list(required = TRUE),
+      Param = list(required = TRUE),
+      Value = list(required = TRUE),
+      NewVar = list(required = TRUE)
+    )
+  )
+  expect_message(
```
jonthegeek, line 87: (Part of an ongoing conversation, I think)

Shouldn't this throw an error? If the columns are required and missing, isn't that an error? If not... what does "required" mean?

I think the message is wrong, too, but @lauramaxwell is probably working on that.

jonthegeek (commented):
Getting close! Just a couple suggested tweaks left.

tests/testthat/test-util-checkSpec.R
jonthegeek: Remaining piece of this is addressed in #1678, we can deal with it later.

lauramaxwell: yep - fixed this in checkSpec to get the message to match correctly. not sure how that got lost

inst/workflow/metrics/kri0002.yaml
```diff
@@ -26,9 +30,10 @@ steps:
   - name: Input_Rate
```
jwildfire, line 30: Let's remove strFilterNumerator and just use additional RunQuery Steps here instead

jonthegeek (approved):
Everything looks good to me! I think you have an unresolved comment or two to confirm, @lauramaxwell , but I'm PRETTY sure this is done!

inst/workflow/metrics/cou0002.yaml
lauramaxwell: yep!

lauramaxwell (approved):

PR #1785, opened 2024-08-15
Mapping workflow refactor

jwildfire:
## Overview
WIP

## Test Notes/Sample Code

Notes:

### DIFF
--- a/inst/workflow/data_mapping.yaml
+++ b/inst/workflow/data_mapping.yaml
@@ -1,141 +1,54 @@
 meta:
   File: mapping.yaml
   description: Data Mappings to prepare for metric calculations
-spec:
-    Raw_SUBJ:
-      subjectid:
-        required: true
-      enrollyn:
-        required: true
-    Raw_AE:
-      subjectid:
-        required: true
-      aeser:
-        required: true
-    Raw_PD:
-      subjectenrollmentnumber:
-        required: true
-      deemedimportant:
-        required: true
-    Raw_LB:
-      toxgrg_nsv:
-        required: true
-    Raw_STUDCOMP:
-      compyn:
-        required: true
-    Raw_SDRGCOMP:
-      sdrgyn:
-        required: true
-      phase:
-        required: true
-    Raw_QUERY:
-      subjectname:
-        required: true
-      querystatus:
-        required: true
-      queryage:
-        required: true
-    Raw_DATACHG:
-      subjectname:
-        required: true
-      n_changes:
-       required: true
-    Raw_DATAENT:
-      subjectname:
-        required: true
-      data_entry_lag:
-        required: true
-    Raw_ENROLL:
-      enrollyn:
-        required: true
 steps:
   - name: RunQuery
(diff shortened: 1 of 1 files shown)


inst/workflow/data_mapping.yaml
```diff
@@ -1,141 +1,54 @@
 meta:
   File: mapping.yaml
   description: Data Mappings to prepare for metric calculations
-spec:
```
lauramaxwell, line 4: So, we are no longer putting any specs/requirements on the raw data? This spec section seems like it could be a useful doc for specifying and updating data requirements from Raw+. Now, all raw data requirements are inside the queries below. I may have missed a discussion, and I don't have strong preferences on this, but it seems a bit more obfuscated for the dev systems team to access the requirements in this way.

inst/workflow/metrics/kri0004.yaml
```diff
@@ -30,9 +30,10 @@ steps:
   - name: Input_Rate
     output: Analysis_Input
     params:
-      dfSubjects: Mapped_Enrolled
-      dfNumerator: Mapped_ImportantPD
-      dfDenominator: Mapped_Enrolled
+      dfSubjects: Mapped_ENROLL
+      dfNumerator: Mapped_PR
```
samussiah, line 34:
```suggestion
      dfNumerator: Mapped_PD
```

inst/workflow/metrics/kri0008.yaml
```diff
@@ -31,9 +31,9 @@ steps:
   - name: Input_Rate
     output: Analysis_Input
     params:
-      dfSubjects: Mapped_Enrolled
-      dfNumerator: Mapped_Query
-      dfDenominator: Mapped_DataChanges
+      dfSubjects: Mapped_ENROLL
+      dfNumerator: Mapped_QUERY
+      dfDenominator: Mapped_DATACHG
```
samussiah, line 36:
```suggestion
      dfDenominator: Mapped_DATACHG
      expFilterNumerator: query_status %in% c('Open','Answered',"Closed")
```

inst/workflow/data_mapping.yaml
samussiah: Totally hear what you're saying - not sure where the full data spec should live, if anywhere.  I like the idea of keeping the metric workflows self-contained, and also avoiding maintaining a spec in two (or more) places.  There's a step missing between the metric workflows and mapping workflow that grabs and collates metric-level specs into a single mapping data spec (also possible y'all have already addressed this).

lauramaxwell: I think the `CombineSpecs()` utility function handles the last part of your comment, if I understand correctly. 

re: a full data spec (including raw data) i think you are right that it shouldn't live in the metrics and that the metrics should have all of the spec information that they need so that they are transportable (assuming the definition of `Mapped_*` data doesn't change).

jwildfire:
Going to go ahead and merge this in to the `mapping-refactor` branch for a consolidated review

PR #1812, opened 2024-09-03
Closes #1722 Study Status Table display bugfix of indiscriminant first five attributes

### LINKED ISSUE #1722: Bugfix: Study status table presents the first five study attributes, regardless of what they are.
## Expected Behavior
In `Report_StudyInfo()`, when hiding the extended information, we want to display protocol ID, nickname, and site/participant numbers.

## Current Behavior
First five, indiscriminately!

## Steps to Reproduce

1.
2.
3.
4.

zdz2101:
## Overview

- Use of `slice(1:5)` contributed to indiscriminant first five attributes
- To fix, need to find some sort of logic that grabs: protocol id, nickname, and site/participant numbers
- Took a look at the `testthat` file that corresponded to `Report_StudyInfo()` to grab what looked like controlled terminology/standard convention names

## Test Notes/Sample Code

Notes:

### DIFF
--- a/R/Report_StudyInfo.R
+++ b/R/Report_StudyInfo.R
@@ -49,7 +49,7 @@ Report_StudyInfo <- function(
     )
 
   show_table <- study_status_table %>%
-    dplyr::filter(Param %in% c("StudyID", "nickname", "enrolled_sites", "enrolled_participants")) %>%
+    dplyr::filter(Param %in% c("GroupID", "nickname", "Status", "SiteCount", "ParticipantCount")) %>%
     dplyr::select("Description", "Value") %>%
     gsm_gt(id = "study_table")
 
--- a/R/Report_StudyInfo.R
+++ b/R/Report_StudyInfo.R
@@ -50,13 +50,13 @@ Report_StudyInfo <- function(
 
   show_table <- study_status_table %>%
     dplyr::filter(Param %in% c("StudyID", "nickname", "enrolled_sites", "enrolled_participants")) %>%
-    gsm_gt(id = "study_table") %>%
-    dplyr::select("Description", "Value")
+    dplyr::select("Description", "Value") %>%
+    gsm_gt(id = "study_table")
 
 
   hide_table <- study_status_table %>%
-    gsm_gt(id = "study_table_hide") %>%
-    dplyr::select("Description", "Value")
+    dplyr::select("Description", "Value") %>%
+    gsm_gt(id = "study_table_hide")
 
   toggle_switch <- glue::glue('<label class="toggle">
   <input class="toggle-checkbox btn-show-details" type="checkbox">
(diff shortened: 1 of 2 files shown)


R/Report_StudyInfo.R
```diff
@@ -40,7 +40,6 @@ Report_StudyInfo <- function(
     Value = unname(unlist(lStudy))
   ) %>%
     dplyr::left_join(dfLabels, by = "Param") %>%
-    dplyr::select("Description", "Value") %>%
```
zdz2101, line 43: With only 2 columns left, is there a way to choose the studyid, nickname, and site/participants numbers just from `Description`? Looking at the testthat file, it looked like `Param` seemed to have some standard conventions/controlled terminology that lends itself to use a `filter` statement but if the descriptions are always consistent, moving this `select` statement isn't necessary and I can modify the `%in%` criteria of the `filter`

zdz2101: @samussiah

tests/testthat/test-Report_StudyInfo.R
```diff
@@ -1,4 +1,5 @@
 lStudy <- list(
+  # Are these variable names standardized?
   StudyID = "Unique Study ID",
```
zdz2101, line 3: Specifically, are these terms in the `lStudy` object always fixed? Or can user put something like `studyid` ?

R/Report_StudyInfo.R
samussiah: What you have looks good!  Better to use `Param` to choose the appropriate rows.  It's not controlled terminology exactly but we can assume it'll have these attributes:

- `GroupID`
- `SiteCount`
- `ParticipantCount`

because each of these columns are explicitly defined in the [`data_reporting` workflow](https://github.com/Gilead-BioStats/gsm/blob/dev/inst/workflow/data_reporting.yaml).

tests/testthat/test-Report_StudyInfo.R
samussiah: The only fixed terms are defined in the [data_reporting workflow](https://github.com/Gilead-BioStats/gsm/blob/dev/inst/workflow/data_reporting.yaml), but I like the idea of applying a filter to additional terms that may or may not exist.

R/Report_StudyInfo.R
```diff
@@ -50,10 +49,13 @@ Report_StudyInfo <- function(
     )
 
   show_table <- study_status_table %>%
-    dplyr::slice(1:5) %>%
+    dplyr::filter(Param %in% c("StudyID", "nickname", "enrolled_sites", "enrolled_participants")) %>%
```
samussiah, line 52:
```suggestion
    dplyr::filter(Param %in% c("GroupID", "nickname", "Status", "SiteCount", "ParticipantCount")) %>%
```

tests/testthat/test-Report_StudyInfo.R
samussiah: These attribute names are Gilead-specific, and they match what comes out of `clindata::ctms_study`.

samussiah (changes requested):
This looks good @zdz2101!  Filtering on `Param` makes total sense and if any study attribute isn't present the function won't error out.  I left one tweak to the subset but otherwise this is ready to go.

samussiah (approved):
💯

PR #1814, opened 2024-09-04
Add type validation to `CheckSpec()`

### LINKED ISSUE #1810: Feature: Check data type in `CheckSpec()`
## Feature Details
use the `type` field in workflow `spec` to check that data matches expected type in `CheckSpec()` function. `type` should be optional, so only perform this check if it exists.

lauramaxwell:
## Overview
Added a simple type validation to `CheckSpec()` to ensure the data is coming in in the correct format based on optional `type` field in `spec`. Also added test to ensure we get an error message when there is a format mismatch

closes #1810 

## Test Notes/Sample Code

Notes:

### DIFF
--- a/R/util-checkSpec.R
+++ b/R/util-checkSpec.R
@@ -53,7 +53,24 @@ CheckSpec <- function(lData, lSpec) {
   # Check that all required columns in the spec are present in the data
   allCols <- c()
   missingCols <- c()
+  wrongType <- c()
   for (strDataFrame in lSpecDataFrames) {
+    #check modes in data
+    imap(lSpec[[strDataFrame]], \(x, idx) {
+      if (!is.null(x$type)) {
+        #check if data is the expected mode
+        res <- x$type %in% mode(lData[[strDataFrame]][[idx]])
+        if (!res) {
+          wrongType <- c(wrongType, idx)
+        }
+      }
+    })
+    if (length(wrongType) > 0) {
+      cli::cli_alert_danger("Not all columns of {strDataFrame} in the spec are in the expected format, improperly formatted columns are: {wrongType}")
+    } else {
+      cli::cli_alert("All specified columns in {strDataFrame} are in the expected format")
+    }
+    #check that required exist in data
     lSpecColumns <- which(sapply(lSpec[[strDataFrame]], function(x) x$required)) %>% names
     lDataColumns <- names(lData[[strDataFrame]])
     allCols <- c(allCols, paste(strDataFrame, lSpecColumns, sep = "$"))
--- a/R/util-checkSpec.R
+++ b/R/util-checkSpec.R
@@ -54,18 +54,25 @@ CheckSpec <- function(lData, lSpec) {
   allCols <- c()
   missingCols <- c()
   for (strDataFrame in lSpecDataFrames) {
-    wrongType <- c()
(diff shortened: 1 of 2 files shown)


R/util-checkSpec.R
```diff
@@ -54,6 +54,30 @@ CheckSpec <- function(lData, lSpec) {
...
+            so_far <- c(so_far, idx)
+          }
+          return(so_far)
+        }
+      },
+      .init = character()
+    )
+
+    if (length(wrongType)) {
+      cli::cli_alert_danger("Not all columns of {strDataFrame} in the spec are in the expected format, improperly formatted columns are: {wrongType}")
```
jonthegeek, line 76: I really feel like these should be `cli::cli_abort()` (see #1678), so we don't keep processing when we already know that it fails the check (`cli_alert_danger()` tells the user that something is wrong but doesn't stop the process). If we did that, we could/should put the required column checks before the type checks.

R/util-checkSpec.R
```diff
@@ -54,6 +54,30 @@ CheckSpec <- function(lData, lSpec) {
...
   for (strDataFrame in lSpecDataFrames) {
+    chrDataFrameColnames <- colnames(lData[[strDataFrame]])
+    #check modes in data
+    wrongType <- purrr::reduce2(
+      lSpec[[strDataFrame]],
+      names(lSpec[[strDataFrame]]),
+      function(so_far, x, idx) {
+        if (!is.null(x$type) && idx %in% chrDataFrameColnames) {
+          #check if data is the expected mode
+          res <- all(x$type %in% mode(lData[[strDataFrame]][[idx]]))
```
jonthegeek, line 65: [Hadley says not to use mode in Advanced R](https://adv-r.hadley.nz/base-types.html#numeric-type:~:text=Do%20not%20use%20these%20functions), but I think it's the right thing here, since we're using "numeric" for type (`typeof()` is more specific about type). I'm not sure this code is exactly right yet. Will x$type always be length 1? Because I'm almost certain `mode()` is always exactly one value. In that case, this should either be either `x$type == mode(...)`, or `any(x$type == mode(...))` if we want to allow x$type to have multiple alternatives.

jonthegeek (approved):
A couple comments, but no blockers. Maya will be excited that we had a real use-case for `reduce()`, though 😊

R/util-checkSpec.R
lauramaxwell: oh yep- should be `==` not `%in%`. this was leftover from when i was trying this out with `class`, which can have multiple values. will change it to `==`

R/util-checkSpec.R
lauramaxwell: yeah, i think this is right. I had this thought as I was writing it, but chose to just match the existing messages instead of challenging status quo🤦‍♀️. so to align with this, we'll need to update all of the existing cli messages in this function but i think i'll go ahead and do that.

jonthegeek: @lauramaxwell I think @jwildfire had thoughts about this/reasons for it to just be the `cli_alert_danger()`, so I'm not sure.

lauramaxwell: i've noted this in #1678 and added that issue to the v2.1.0 milestone. can discuss further if needed, but i think its a separate issue from this PR

jonthegeek: Agreed!

PR #1815, opened 2024-09-04
Closes #1779 Enhance `CombineSpecs()` with deduplication, enforcing `required` and adding checks for `type` 

### LINKED ISSUE #1779: Feature: Refine `CombineSpecs()`
## Feature Details
In cases where we have duplicated column names in specs that we are combining, we will want to follow the following rules:

- de-dupe column names within the same df
- if one spec has `required: true` and the other has `required: false` set the combined spec to `required:true`
- if specs have different `type` the we should use the first type and throw a warning about the mismatch. (`type` is not implemented in specs yet)

zdz2101:
## Overview
`utils-CombineSpecs.R` now contains 3 functions:
1. `CombineSpecs()` which combines a list of specs into a single spec
2. `combine_domain()` which does the deduplification of columns
3. `update_column()` which does the reinforcement of `required = TRUE` if its set to `TRUE` anyhwere else in the spec(s), and throws warning message of `type` is inconsistent

## Test Notes/Sample Code
Added new unit tests in corresponding `testthat` file


lauramaxwell (approved):
looks great to me!

PR #1824, opened 2024-09-05
Closes #1686 Renders a message in `KRI_Report()` when there are 0 rows in the summary table

### LINKED ISSUE #1686: Feature: Render a message in `KRI_Report()` when there are 0 rows in the summary table
## Feature Details

Render a message (instead of an empty table) when there are 0 rows in the summary table. Something like:  "0 of 15 groups selected. Update filter to see more data."
 
image

zdz2101:
## Overview
- The little blurb below the overview table is rendered by `Report_OverviewText()`
- By adding an additional `if()` statement that searches for the 0 rows of flagged entries, you can conditionally print/cat this message

## Test Notes/Sample Code

Notes:

### DIFF
--- a/R/Report_OverviewText.R
+++ b/R/Report_OverviewText.R
@@ -28,6 +28,10 @@ Report_OverviewText <- function(lSetup, dfResults, lStudy) {
     group_type <- "countries"
   }
 
+  if(is_empty(amber_or_red_KRI_groups)) {
+    cat("0 of 15 groups selected. Update filter to see more data.")
+  }
+
   glue("
         <div style = 'margin-top: 1em;'>As of {lSetup$SnapshotDate}, {lSetup$StudyID} has {as.numeric(lStudy$ParticipantCount)} participants enrolled across {lStudy$SiteCount} {group_type}. {length(red_KRI_groups)} Site-KRI combinations have been flagged across {length(unique(red_KRI_groups))} {group_type} as shown in the Study Overview Table above.</div>
        - <div>{length(unique(red_KRI_groups))} {group_type} have at least one red KRI</div>
--- a/inst/report/Report_KRI.Rmd
+++ b/inst/report/Report_KRI.Rmd
@@ -87,12 +87,6 @@ htmltools::div(
 
 ```
 
-```{r echo=FALSE, warning=FALSE, results = 'asis'}
-if(all(dfResults_current$Flag == 0 | is.na(dfResults_current))) {
-  cat("0 of 15 groups selected. Update filter to see more data.")
-}
-```
-
 :::
 
 ```{r, echo=FALSE, results = 'asis'}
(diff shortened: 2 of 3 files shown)


R/Report_OverviewText.R
```diff
@@ -28,6 +28,10 @@ Report_OverviewText <- function(lSetup, dfResults, lStudy) {
     group_type <- "countries"
   }
 
+  if(is_empty(amber_or_red_KRI_groups)) {
+    cat("0 of 15 groups selected. Update filter to see more data.")
```
zdz2101, line 32: Is it always 15? Can possibly turn this into a programmatically generated number based on unique of groups in the respective report

zdz2101 (commented):
@jwildfire

R/Report_OverviewText.R
lauramaxwell: I think this number would be `length(unique(dfResults$GroupID))` because those are all the possible groups that can have flags

lauramaxwell: instead of "selected" might say something like "0 of XX groups applicable to this filter. Update filter to see more data."

lauramaxwell (commented):
After review, it appears that this message actually needs to be in the Study_Overview widget in `rbm.viz`, because the message is dependent on what dropdown filter is selected, and ideally would be placed within the table. I'll move this issue over to `{rbm.viz}` to see if we can solve it there. @samussiah

zdz2101:
> After review, it appears that this message actually needs to be in the Study_Overview widget in `rbm.viz`, because the message is dependent on what dropdown filter is selected, and ideally would be placed within the table. I'll move this issue over to `{rbm.viz}` to see if we can solve it there. @samussiah

Is that still true for the rendering of `inst/report/Report_KRI.Rmd` too?

lauramaxwell:
> Is that still true for the rendering of `inst/report/Report_KRI.Rmd` too?

Yes, I meant to say that it needs to be in the Group Overview widget, not the Study Overview widget, my bad. But, that widget gets called via `Widget_GroupOverview()` in line 79 of `Report_KRI.Rmd`, so updating the widget will make the message appear in the KRI report (and any other reports that use this widget) as well.

lauramaxwell:
@samussiah @jwildfire new issue for this filed in `rbm-viz` here: https://github.com/Gilead-BioStats/rbm-viz/issues/404

not the most urgent thing, but just putting it on your radar

zdz2101:
I see, I suppose I misinterpreted the way the report works/that this is held constant. I figured if there's already always a function ran that renders a text output below the table, it would make sense to put the message there, but that makes sense if the widget is used elsewhere and a text blurb isn't immediately below you would want a message in lieu of the table.

jwildfire:
Agree this needs to be done in `rbm-viz` per the discussion above. Can we just close this PR since this will be resolved via a widget update at some point?

PR #1827, opened 2024-09-06
Add all required spec tables to workflows

### LINKED ISSUE #1819: Feature: Add all required tables to workflow specs
## Feature Details
Convert `spec` from a list to an array and list all required tables. For Reporting and Analysis tables, no column level spec is needed needed. For `Map_*` data specify a list of columns and `required`/`type`

Also requires modification to `checkSpec()` and `combineSpec()`

lauramaxwell:
## Overview
Added `Reporting_*` and `Analysis_*`  tables to specs without requiring column specification. Updated `data_mapping.yaml` to include a spec as well.

update `Check_Spec()` to handle tables without column specs

## Connected Issues

- Closes #1819


lauramaxwell:
@jwildfire and @samussiah after a few tweaks that came out of some conversations re: `gsm.template`, this is ready for review now.

tests/testthat/test-util-checkSpec.R
```diff
@@ -159,3 +159,36 @@ test_that("Validate column type works", {
...
+    df2 = data.frame(x = 7:9, y = 10:12),
+    df3 = data.frame(z = 1:3, t = 20:22)
+  )
+
+  expect_message(
+    expect_message(expect_message(
+      expect_message(CheckSpec(lData, lSpec), "All 3 data"),
+      "All specified"
+    ), "All specified"),
+    "All 4 required"
```
samussiah, line 191: 4 includes `a`, `b`, `x`, and `y`?

samussiah (changes requested):
Looks good @lauramaxwell!  The `Raw_SUBJ` spec should probably include `enrollyn` since it appears in a `WHERE` clause.  Otherwise good to go.

tests/testthat/test-util-checkSpec.R
lauramaxwell: right. at this point, we are only checking columns that specify particular columns, not those that require `_all`. I suppose, I could add those columns to the message count, but it might complicate things a bit.

samussiah (approved):
Looks good!

Issue #1804, opened 2024-08-28
QC: update `pkgdown` yamls to cache dependencies


lauramaxwell:
## QC Details
r dependencies are not currently cached in the pkgdown yamls. This adds ~8 mins to the check. Update the `setup-r-dependencies@v2` step to properly cache the dependencies, so only new dependencies need to be added at each run. Likely, we will need to remove `local::.` from the following code chunk and load the local R package in a separate step.

```
uses: r-lib/actions/setup-r-dependencies@v2
        with:
          extra-packages: any::pkgdown, local::.
```

zdz2101:
`setup-r-dependencies` by default has cache: TRUE, so some sort of cache is being actively saved/used in each of the yaml files, But because each workflow looks to have its own and updates the cache, looks like it can "miss" or not hit the proper cache if those cache-keys in the background don't match or possibly expired since github acions automatically deletes if its >7 days old unaccessed. I tinkered around with setting cache as FALSE and restoring a universal cache, see  [the yaml that creates the cache](https://github.com/Gilead-BioStats/gsm/blob/fix-1804/.github/workflows/cache-pkg.yaml) and [the respective action](https://github.com/Gilead-BioStats/gsm/actions/runs/11021676284) that ran on it. 

May be worth cleaning up the cache anyway too, repo is getting close to allowable limit

jonthegeek:
> May be worth cleaning up the cache anyway too, repo is getting close to allowable limit

GitHub auto prunes the least-recently-used cache items, so we shouldn't need to manually prune.

jonthegeek:
Huh, the r-lib action is smarter about caching than I realized:
```
          key: ${{ format('{0}-{1}-{2}-{3}', steps.install.outputs.os-version, steps.install.outputs.r-version, inputs.cache-version, hashFiles(format('{0}/.github/pkg.lock', inputs.working-directory ))) }}
          restore-keys: ${{ format('{0}-{1}-{2}-', steps.install.outputs.os-version, steps.install.outputs.r-version, inputs.cache-version) }}
```

What that means: it creates a key based on the OS, R version, cache version, and a hash of the repo's working dir (and pkg lock file), and restores based on OS, R version, and cache version. I think I overrode it in this YAML thinking it was making a cache specific to the PR (which some other actions do by default), which wasn't very helpful.

[...]

PR #1828, opened 2024-09-09
Update sample reporting data

### LINKED ISSUE #1703: Feature: Add in all KRIs to `reportingResults`
## Feature Details

Add longtudinal data for KRI0003-KRI0012 and then update `reportingMetrics`, `reportingBounds` of reporting data model as needed.

### LINKED ISSUE #1793: QC: Add sample data generation scripts to `data_raw/`
## QC Details
To increase transparency and reproducibility, add scripts that generate the sample data frame CSVs

### LINKED ISSUE #1800: QC: Update sample reports to include longitudinal data
## QC Details
Currently just using one snapshot, update to use Reporting model

lauramaxwell:
## Overview

Update sample reporting data to: 
1. be longitudinal
2. include all KRIs
3. be available for site and country reporting
4. Include data simulation script in `data-raw` directory

This new longitudinal reporting data will be used in the sample reports in `pkgdown`, which addresses #1703.

**Note**: The site-level reporting data has no suffix, but country-level reporting data has `_country` suffix.

## Connected Issues

- Closes #1800 
- Closes #1793
- Closes #1703


.github/workflows/pkgdown-dev2main.yaml
```diff
@@ -33,88 +33,40 @@ jobs:
       - name: Generate sample reports
         shell: Rscript {0}
         run: |
-          devtools::load_all()
+          devtools::load_all(dependencies = T)
```
jonthegeek, line 36: `pkgload::load_all()` (called via `devtools::load_all()` doesn't have a `dependencies` argument. This should throw an error. Were you thinking of `devtools::install()`?

.github/workflows/pkgdown.yaml
```diff
@@ -32,90 +32,43 @@ jobs:
       - name: Generate sample reports
         shell: Rscript {0}
         run: |
-          devtools::load_all()
+          devtools::load_all(dependencies = T)
```
jonthegeek, line 35:
```suggestion
          devtools::load_all()
```
See above.

jonthegeek (approved):
This looks great! Unfortunately... it won't work for gsm.app. I need us to export *all* of the tables (the datasets used to construct the reports), and that doesn't make sense in gsm. Can/should we make a gsm.sampledata package that owns the input data (the stuff you randomly generated here)? It's generally considered better to separate large datasets from the "main" package so you don't have to keep archiving those datasets with every release, otherwise I'd ask that we include it all in GSM.

I guess technically I could regenerate the data (since you include a seed), but it doesn't really make sense for it to be "inside" gsm.app, either.

.github/workflows/pkgdown-dev2main.yaml
lauramaxwell: oh whoops- that can actually just go away- i was thinking about addressing #1804, but then stopped part-way through and forgot to get rid of all the evidence. but yeah- i was probably thinking of `install` here.

jonthegeek (approved):
That should do it!

PR #1833, opened 2024-09-11
Closes #1636 allows `Visualize_Metric()` to run on just results

### LINKED ISSUE #1636: Feature: Make parameters optional in `Visualize_Metric()`
## Feature Details
Visualizations should be able to run with just `dfResults`. All other parameters and data frames should be optional.

zdz2101:
## Overview
- `Widget_Barchart()` , `Widget_Scatterplot()`, `Widget_TimeSeries()` should be able to run on just results
- Logic was introduced with the use of insertion of `NULL` and `if(is.null())` to allow `Visualize_Metric()` to run on just the results as well


R/Visualize_Metric.R
```diff
@@ -133,12 +141,16 @@ Visualize_Metric <- function(
...
       vThreshold = vThreshold
     )
-
-    lCharts$metricTable <- Report_MetricTable(
-      dfResults = dfResults_latest,
-      dfGroups = dfGroups,
-      strGroupLevel = lMetric$GroupLevel
-    )
+    if(!is.null(dfGroups)) {
+      lCharts$metricTable <- Report_MetricTable(
```
zdz2101, line 145: @lauramaxwell do you know if `Report_MetricTable()` is meant to be one of the visualizations that can be rendered with just results? If so, I think I need to make some modifications there

samussiah: Ideally, although we should always have participant counts.  Still if it's not too much effort this is a good candidate to optionalize group metadata.

R/Report_MetricTable.R
```diff
@@ -28,17 +28,20 @@
 #' @export
 Report_MetricTable <- function(
   dfResults,
-  dfGroups,
+  dfGroups = NULL,
   strGroupLevel = c("Site", "Country", "Study"),
   strGroupDetailsParams = NULL,
   vFlags = c(-2, -1, 1, 2)
 ) {
+  if(!is.null(dfGroups)) {
```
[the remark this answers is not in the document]
samussiah: This is exactly what I had in mind!

samussiah (changes requested):
Almost there @zdz2101 - `Report_MetricTable` looks great, but if `dfBounds` or `dfMetrics` aren't passed into `Visualize_Metric` the function bombs where it filters on `strMetricID`.  btw I added a little logic to the individual widgets to more robustly handle the data type of `lMetric`, so make sure you pull the latest commit.

Issue #1830, opened 2024-09-09
Feature: Parse `params` list in `lStep`


lauramaxwell:
## Feature Details
Per the gsmExtensions vignette, we want to parse the `params` list, so that each item is split in to `paramName` and `paramVal`. For example, `lData: Mapped_AE` is parsed to `paramName = 'lData'` and `paramVal = 'Mapped_AE'` and the following rules are applied in order: 

1. If `paramVal` is equal to "lMeta", the the full `lMeta` object is passed to the function (for the given `paramName`). 
2. If `paramVal` is equal to  "lData", the full `lData` object is passed to the function. 
3. If `paramVal` is found in `names(lMeta)`, that property is pulled from `lMeta` (e.g. `lMeta${paramVal}`) and passed to the function.
4. If `paramVal` is found in `names(lData)`, that property is pulled from `lData` (e.g. `lData${paramVal}`) and passed to the function.
5. Otherwise `paramVal` is passed to the function as a string.

PR #1841, opened 2024-09-13
Refactor Data Ingestion Workflow

### LINKED ISSUE #1840: Feature: Add option to import data in RunWorkflow
## Feature Details

Add an optional `lInputConfig` parameter to `RunWorkflow` that allows users to specify data locations and call `LoadData()` and `SaveData()` functions instead of passing and returning `lData` directly.

jwildfire:
## Overview
Draft PR to refactor default data pipeline for improved data ingestion. 

Fixes #1840 #1830 and probably others ... 

## Connected Issues

- Closes #XXX


R/util-SaveData.R
```diff
@@ -0,0 +1,47 @@
+SaveData <- function(lData, lInputConfig) {
```
zdz2101, line 1: @lauramaxwell was this meant to be exported/roxygened?

R/RunStep.R
```diff
@@ -4,18 +4,38 @@
...
+#'
+#' @param lStep `list` single workflow step (typically pulled from `lWorkflow$steps`). Should
+#'   include the name of the function to run (`lStep$name`), name of the object where the function result should be saved (`lStep$output`) and configurable parameters (`lStep$params`) (if any)
 #' @param lData `list` a named list of domain level data frames.
+#' @param lSpec `list` a data specification containing required columns. See `vignette("gsm_extensions")`.
 #' @param lMeta `list` a named list of meta data.
 #'
 #' @examples
-#' wf_mapping <- MakeWorkflowList("mapping")
+#' wf_mapping <- MakeWorkflowList(strPath = "/inst/workflow/1_mappings")
```
zdz2101, line 38: this example's output is slightly changed I think it needs some modification 
image

R/util-SaveData.R
lauramaxwell: yeah, we should have a roxygen header added here- it's only really applicable if we are using data from an external/cloud source, but will definitely be useful once we get that all set up, so it should be exported.

R/RunStep.R
lauramaxwell: hmmm, yeah. since we are just running the `ae` metric with `RunStep`, we can probably just run the `1_mappings/AE.yaml` workflow to create the data. The name of the lData dfs will need to be updated to match what is in that `AE.yaml`, but that should get you closer to a working example of `RunStep()`

zdz2101: I suppose I was thrown off by the `No spec found in workflow. Proceeding without checking data.` statement, but looks like that is being introduced in this current version of `dev` that does not currently exist on `main` . Otherwise this example runs "the same" as the preexisting one on `main` just new messaging, but the outcome is actually the same.

R/util-SaveData.R
zdz2101: Do we want this done for the upcoming release? Applies to `LoadData` as well then I suppose

inst/workflow/2_metrics/cou0001.yaml
```diff
@@ -62,4 +62,12 @@ steps:
     params:
       dfFlagged: Analysis_Flagged
       nMinDenominator: nMinDenominator
-
+  - name: list
+    output: cou0001
+    params:
+      id: ID
+      input: Analysis_Input
```
lauramaxwell, line 69: @jwildfire was it intentional to change all of the `Analysis_*` outputs to remove that prefix and make them lowercase? It seems like we are keeping `Raw_*` and `Reporting_*`, so I wasn't sure why we would want to remove the `Analysis_` prefix for those data objects.

R/util-SaveData.R
jwildfire: Yes, let's file an issue and assign to @samussiah. Need to make sure this works with a standard `lConfig` object.

R/RunStep.R
jwildfire: Yeah, I didn't tweak the examples at all. Feel free to modify/simplify as needed.

inst/workflow/2_metrics/cou0001.yaml
lauramaxwell: okay, i'll keep it as is, but @samussiah - let me know if you have thoughts/preferences and we can log a quick issue to change the naming

lauramaxwell (approved):
We're getting there! This is looking good- I'm merging this in so that we can get to work on the remaining issues for the v2.1 release 🧑‍💻

PR #1843, opened 2024-09-13
MakeMetricTable()

### LINKED ISSUE #1835: Feature: Data_MetricTable
## Feature Details
As the lead {gsm.app} developer, in order to utilize gsm outputs without unnecessary constraints, I would like to process dfResults + dfGroups + strGroupLevel into a data.frame, that I can then display in a standardized way.

## Possible Implementation
`Report_MetricTable()` has these steps at the end:

```r
 %>% 
    kableExtra::kbl(format = "html", escape = FALSE) %>% 
    kableExtra::kable_styling("striped", full_width = FALSE)
```

I want everything above that, but I don't want that format for the output. I want to be able to standardize on whichever display is best for shiny, and use that across all of my data. That will also let me work out the input updates across the app one and use the same strategy everywhere, rather than having to sort it out across several different output formats.

## Additional Comments
This is similar to #1820. We should probably discuss whether this is the approach we want to take for all of these.

jonthegeek:
## Overview
Split MakeMetricTable out from Report_MetricTable.

## Connected Issues

- Closes #1835


jwildfire (commented):
Looks good to me.

lauramaxwell (approved):
I like it!

PR #1850, opened 2024-09-18
Closes #1653 adds Flag Over Time Widget to Study Overview as a tab

### LINKED ISSUE #1653: Feature: Add `FlagOverTime()` table to report when more than 1 snapshot is included
## Feature Details

Render the FlagOverTime table in the report. I think the simplest approach for now is to add this as a new tab under the Study Overview section. 

We can consider a more complex implementation in a future release. For example, users could drill down from the "Overview" table to the "FlagOverTime" table; when a user clicks a row for a site in the current overview table, the FlagOverTime table renders for that row.  Would have to think about implementation details ... could hack something together in the report, or do a more robust implementation in `rbm-viz`

cc: @samussiah @jonthegeek @lauramaxwell

zdz2101:
## Overview
Mimic the `Visualize_Metric()` / `Report_MetricCharts()` and like-kind that generate the tabbed portions of Results section in the KRI report but for:
Group Overview and Flag Over Time under Study Overview


zdz2101:
image

image

Tabulated form of Study Overview is ready-ish, or certainly a working rough draft is available. Couple points to raise though:
1. For the Flag Over Time visualization, it seems that this can blow up in size/ability to render (see top image orange circle to give an idea how much scrolling may be done), as there are up to (m * n * p) boxes, where m is number of sites, n is number of kri's and p is number of timepoints (see bottom image)
2.  I was actually unable to run example 3.3 Site-Level KRI Report locally, that I had to use the filtered snapshot date for the top image to check the tabs actually work and just `Widget_FlagOverTime()` to get the bottom image, obviously this is will dependent on hardware capabilities or whatever is running this report
3.  It may be more useful to limit what is actually displayed in the Flag Over Time visualization, perhaps display `New Flag?` rows only? Further discussion needed

@lauramaxwell @jwildfire

lauramaxwell:
We could also potentially filter on "Any flag in most recent period", which would include new flags and unresolved flags. This might be a question for the user group, to ensure we meet their requirements.

zdz2101:
Shaving it down to not-all-green rows certainly did the trick for rendering on my end locally 
image

inst/report/Report_KRI.Rmd
```diff
@@ -74,17 +74,33 @@ if(setup$GroupLevel == "Site"){
...
+GroupOverViz <- Widget_GroupOverview(
+  dfResults = dfResults_current,
+  dfGroups = params$dfGroups,
+  dfMetrics = params$dfMetrics, 
+  strGroupLevel = setup$GroupLevel,
+  strGroupLabelKey = GroupLabelKey
 )
 
+# Remove sites whose metrics have no flags 
+dfResults_atleastoneflag <- params$dfResults %>%
```
zdz2101, line 86: @lauramaxwell @jwildfire Do we want to turn this snippet into some sort of utility function of some sort that filters out the no-flag rows? Wasn't sure how much data manipulation was appropriate, if any, inside the `Report_KRI.Rmd` was

inst/report/Report_KRI.Rmd
```diff
@@ -74,17 +74,33 @@ if(setup$GroupLevel == "Site"){
...
+FlagOverViz <- Widget_FlagOverTime(
+  dfResults = dfResults_atleastoneflag,
+  dfMetrics = params$dfMetrics,
+  strGroupLevel = setup$GroupLevel,
+  strFootnote = "*Only sites with at least one flag are shown."
+)
+
+overview_viz <- list(groupOverviewJS = GroupOverViz, flagOverTimeJS = FlagOverViz)
+
+Report_MetricCharts(overview_viz, overview = TRUE)
```
zdz2101, line 103: Felt easier to make a small modification to `Report_MetricCharts()` that does the tabbing/knitting/printing than create a new function or hand write it into here? Certainly can modify

inst/report/Report_KRI.Rmd
lauramaxwell: @zdz2101 yes, a utility function to handle this filter would be best. Want to minimize custom data manipulation inside the Rmd when possible.

R/Report_MetricCharts.R
```diff
@@ -34,7 +46,9 @@ Report_MetricCharts <- function(lCharts, strMetricID = "") {
       barMetricJS = paste0(fontawesome::fa("chart-simple", fill = "#337ab7"), "  KRI Metric"),
       timeSeriesContinuousScoreJS = paste0(fontawesome::fa("chart-line", fill = "#337ab7"), "  KRI Score"),
       timeSeriesContinuousMetricJS = paste0(fontawesome::fa("chart-line", fill = "#337ab7"), "  KRI Metric"),
-      timeSeriesContinuousNumeratorJS = paste0(fontawesome::fa("chart-line", fill = "#337ab7"), "  Numerator")
+      timeSeriesContinuousNumeratorJS = paste0(fontawesome::fa("chart-line", fill = "#337ab7"), "  Numerator"),
+      groupOverviewJS = paste0(fontawesome::fa("table", fill = "#337ab7"), "  Group Overview"),
+      flagOverTimeJS = paste0(fontawesome::fa("table", fill = "#337ab7"), "  Flag Over Time")
```
lauramaxwell, line 51: Matching tab label with chart label
```suggestion
      flagOverTimeJS = paste0(fontawesome::fa("table", fill = "#337ab7"), "  Flags Over Time")
```

inst/report/Report_KRI.Rmd
```diff
@@ -74,17 +74,27 @@ if(setup$GroupLevel == "Site"){
...
+  dfGroups = params$dfGroups,
+  dfMetrics = params$dfMetrics, 
+  strGroupLevel = setup$GroupLevel,
+  strGroupLabelKey = GroupLabelKey
+)
+
+# Remove sites whose metrics have no flags at all timepoints
+dfResults_flags <- FilterByFlags(params$dfResults)
+
+FlagOverViz <- Widget_FlagOverTime(
```
lauramaxwell, line 88: Per issue #1653, we only want to include this chart when the number of unique snapshot dates in dfResults > 1, so adding some logic in `Report_KRI.Rmd` would help achieve that.

lauramaxwell (changes requested):
A few small comments, but overall, this is looking good and load time wasn't too bad

lauramaxwell (approved):
Working great for me! 👍

PR #1861, opened 2024-09-25
Run `{gsm}` using local data store.

samussiah:
## Overview
This branch enables generating a snapshot using a local data store. There be dragons:

- Might have broken in-memory workflows...
- Need to adhere to either `Analysis_*` or `Metric_*`; right now it's a mixture of the two.
- Added explicit support for `save: true` in workflows to keep with the one worklow / one output paradigm while also enabling storage of intermediate outputs like `Analysis_Input`.
- File format limited to .csv currently.
- Moved `ApplySpec` out of `Ingest` and into its own function for use in `LoadData`, where data types are character by default.

## Test Notes/Sample Code
Run `inst/examples/AA-AA-000-0000/scripts/run-snapshot.R` to generate a snapshot.

## Connected Issues
- Closes #1818 
- Closes #1834 
- Closes #1854

PR #1864, opened 2024-09-30
Load/Save - Take 2

jwildfire:
## Overview

Redo of #1861 

## Connected Issues

- Closes #XXX


inst/examples/AA-AA-000-0000/config/2024-09-26/config.yaml
```diff
@@ -0,0 +1,258 @@
...
+  clindata: clindata
+  local: ./data
+  s3: s3://AA-AA-000-0000/data
+schemas:
+  Raw:
+    db:
+      clindata: clindata
+    domains:
+      Raw_STUDY:
+        db: clindata
```
jwildfire, line 12: wonder if we can simplify just using the info for the current study: 

```
schemas: 
   Raw: 
      db: clindata
      domains: 
          - Raw_Study: ctms_study
          - Raw_Site: ctms_site
          - ...
 ```

File: inst/examples/AA-AA-000-0000/config/2024-09-26/config.yaml [hunk 2]
```diff
@@ -0,0 +1,258 @@
...
+      Raw_DATACHG:
+        db: clindata
+        table: edc_data_points
+      Raw_QUERY:
+        db: clindata
+        table: edc_queries
+  Mapped:
+    db:
+      local: ./data/2024-09-26/Mapped/{ID}.csv
+      s3: s3://AA-AA-000-0000/data/2024-09-26/Mapped/{ID}.parquet
```
jwildfire, line 50: probably just need one of these for any given study, right?

inst/examples/AA-AA-000-0000/config/2024-09-26/config.yaml
```diff
@@ -0,0 +1,258 @@
...
+- id: report_kri_site
+  repo: Gilead-BioStats/gsm
+  path: inst/workflow/reports/report_kri_site.yaml
+  version: dev
+  package: gsm
+  file: workflow/reports/report_kri_site.yaml
+metrics:
+- id: kri0001
+  repo: Gilead-BioStats/gsm
+  path: inst/workflow/metrics/kri0001.yaml
```
jwildfire, line 142: I wonder if we really need this. Feel like we could just have almost all of this in the workflow yaml header. And then just use standard paths ... Maybe add it in the future if we need customization

inst/examples/AA-AA-000-0000/config/2024-09-26/config.yaml
```diff
@@ -0,0 +1,258 @@
...
+  package: gsm
+  file: workflow/metrics/kri0003.yaml
+- id: kri0004
+  repo: Gilead-BioStats/gsm
+  path: inst/workflow/metrics/kri0004.yaml
+  version: dev
+  package: gsm
+  file: workflow/metrics/kri0004.yaml
+SnapshotDate: '2024-09-26'
+domains:
```
jwildfire, line 165: `domains` feels duplicative to some of the stuff above ...

jwildfire (commented):
I think this is close, but needs a few more tweaks - though maybe most can be done in gsm.template

R/RunWorkflow.R
```diff
@@ -90,6 +94,14 @@ RunWorkflow <- function(
 
     lWorkflow$lData[[step$output]] <- result
     lWorkflow$lResult <- result
+
+    if (!is.null(step$save) && step$save) {
```
jwildfire, line 98: I think we should get rid of step$save. 

Instead, I think we can probably support saving a list of data by adding some basic iteration in SaveData(). Something along the lines of: 

```
if(is.list(lWorkflow$result)){
   lWorkflow$result %>% map(SaveData)
} else if(is.data.frame(lWorkflowResult)) {
 #existing code goes here
} else {
    cli_warning("data format not supported, not saving")
```

lauramaxwell: So does this mean we are putting the last step of the metric yamls back in- where we write all tables out as a list? In the current version, this step has been removed.

lauramaxwell: Alternatively, we just set `bReturnResult` to FALSE when running locally and keep the `save` field, but add some logic so that it only triggers a `SaveData()` when `lConfig` is not null.

samussiah:
@jwildfire, @lauramaxwell - load/save has been removed from this branch so it's ready to review.

lauramaxwell:
tests and checks are passing now! i think everything looks good, but i'll give it another look before merging.

lauramaxwell (approved):
This looks good!

PR #1891, opened 2024-10-15
Only show flag-rows in Report_FlagOverTime()

### LINKED ISSUE #1879: Feature: Reduce render time of Flags Over Time module in report.
## Feature Details
With more and more snapshots the flags over time table takes a long time to render. Would be good to figure out how to speed up the process.

jonthegeek:
## Overview
If a metric has never been flagged at a given site, do not show that metric at that site. This might not be desirable, but it's the best option I found.

## Connected Issues

- Closes #1879.

### DIFF
--- a/R/Report_FlagOverTime.R
+++ b/R/Report_FlagOverTime.R
@@ -28,6 +28,9 @@ Report_FlagOverTime <- function(
   )
 
   dfFlagOverTime %>%
+    dplyr::filter(
+      dplyr::if_any(dplyr::all_of(date_cols), ~!is.na(.x) & .x != 0)
+    ) %>%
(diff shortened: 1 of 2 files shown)


lauramaxwell:
@zdz2101 - isn't this the behavior you implemented in the reporting output?

zdz2101:
> @zdz2101 - isn't this the behavior you implemented in the reporting output?

Yes and no, I addressed it specifically at the level of `inst/report/Report_KRI.Rmd` by filtering the data that is put into the widget, instead of modifiying the widget itself

jonthegeek:
Getting rid of background coloration DRAMATICALLY sped this up. I don't think we need this PR. See next PR.

PR #1908, opened 2024-10-23
Fix country overview text

lauramaxwell:
## Overview

Update Report_OverviewText and Report_Setup to handle country overview text properly. also deals with two existing bugs that displayed only red flag counts in the description and calculated `no_alert_groups` incorrectly.

Previous text:
image

Updated text:
image

## Connected Issues

- Closes #XXX


jonthegeek (approved):
Looks great as long as tests pass!

PR #1913, opened 2024-10-25
updated class structure

### LINKED ISSUE #1912: Feature: Use "name__class" method in `sqldf()`
## Feature Details

`RunQuery()` uses `sqldf()` to perform its SQL queries, but it currently does not enforce classes when returning data. Using the "name__class" method allows the user to specify the class of the column in the query, and `sqldf` will return the data in the proper class. 

When a `spec` entry contains "type", we should use that field to ensure class compliance throughout the mapping process.

andersonjake1988:
## Overview
added `method = "name__class"` to `sqldf` in `RunQuery` and adjusted `ApplySpec` to account for adjustment

## Connected Issues

- Closes #1912

### DIFF
--- a/R/util-ApplySpec.R
+++ b/R/util-ApplySpec.R
@@ -38,13 +38,8 @@ ApplySpec <- function(dfSource, columnSpecs, domain) {
   columnMapping <- columnSpecs %>% imap(
     function(spec, name) {
       mapping <- list(target = name)
-
-      if ("source_col" %in% names(spec)) {
-        mapping$source <- spec$source_col
-      } else {
-        mapping$source <- name
-      }
-
+      mapping$source <- spec$source_col %||% name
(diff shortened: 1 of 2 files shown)


R/util-ApplySpec.R
```diff
@@ -38,13 +38,8 @@ ApplySpec <- function(dfSource, columnSpecs, domain) {
...
     function(spec, name) {
       mapping <- list(target = name)
-
-      if ("source_col" %in% names(spec)) {
-        mapping$source <- spec$source_col
-      } else {
-        mapping$source <- name
-      }
-
+      mapping$source <- spec$source_col %||% name
```
lauramaxwell, line 41: nice simplification!

lauramaxwell (approved):
It's working well for me!

PR #1923, opened 2024-10-30
Fix Ingest return value

### LINKED ISSUE #1924: Bugfix: Optional Columns in 1st Combined Spec
## Expected Behavior
If I try to combine two specs, and 1 has optional columns, there isn't a problem.

## Current Behavior
If the first spec has a df with an optional column, CombineSpec throws an error. If the *second* spec has an optional column that isn't in the first step, it seems to work, but I suspect it doesn't really.

## Possible Solution
Fixed, PRing in a sec.

## Steps to Reproduce
```r
spec1 <- list(
  df1 = list(
    col1 = list(required = TRUE),
    col2 = list(required = FALSE)
  )
)

spec2 <- list(
  df1 = list(
    col3 = list(required = TRUE)
  )
)

combined <- CombineSpecs(list(spec1, spec2), bIsWorkflow = FALSE)
```

jonthegeek:
## Overview
Fix Ingest return value (spotted by actually using the docs 😊).
Update `CombineSpecs()` to better handle non-required columns.

## Test Notes/Sample Code
Added tests to make sure this doesn't regress.

## Connected Issues

- Closes #1924

### DIFF
--- a/R/util-ApplySpec.R
+++ b/R/util-ApplySpec.R
@@ -35,16 +35,20 @@ ApplySpec <- function(dfSource, columnSpecs, domain) {
   }
 
   # write a query to select the columns from the source
-  columnMapping <- columnSpecs %>% imap(
-    function(spec, name) {
-      mapping <- list(target = name)
-      mapping$source <- spec$source_col %||% name
-      mapping$type <- spec$type %||% NULL
-      return(mapping)
-    }
-  )
+  columnMapping <- columnSpecs %>%
+    imap(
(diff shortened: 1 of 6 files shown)


jonthegeek:
Fixing #1924, 1 sec.

jonthegeek:
Making Ingest() friendlier to optional columns.

samussiah (approved):
Logic checks out, and kudos on the unit test scenarios!

PR #1930, opened 2024-11-04
Update `Metrics.yaml` to not generate `lWorkflows` within steps

### LINKED ISSUE #1929: Feature: Make `Metrics.yaml` based on lWorkflows input
## Feature Details

Currently, `Metrics.yaml` creates the `lWorkflows` object based on `gsm`, but this should work based on whatever `lWorkflows` object the user needs for a particular report (i.e. from a different package/extension). So, this yaml should just take `lWorkflows` as an input from the `lData` list in `RunWorkflows`.

lauramaxwell:
## Overview

`lWorkflows` should be based on the current module/study being run, not always just the `gsm` workflows. This simply updates the flow to take an `lWorkflows` argument in the `lData` list in `RunWorkflows()` to run Metrics.yaml properly. 

Note: `lWorkflows` was already being fed into `RunWorkflows()` in this way, so this is not an update to process, just a simplification.

## Connected Issues

- Closes #1929

### DIFF
--- a/inst/examples/3_ReportingWorkflow.R
+++ b/inst/examples/3_ReportingWorkflow.R
@@ -55,7 +55,8 @@ analyzed <- RunWorkflows(metrics_wf, mapped)
 
 # Step 3 - Create Reporting Layer - create reports using metrics data
 reporting_wf <- MakeWorkflowList(strPath = "workflow/3_reporting")
-reporting <- RunWorkflows(reporting_wf, c(mapped, list(lAnalyzed = analyzed, lWorkflows = metrics_wf)))
+reporting <- RunWorkflows(reporting_wf, c(mapped, list(lAnalyzed = analyzed,
+                                                       lWorkflows = metrics_wf)))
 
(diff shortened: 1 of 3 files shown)


zdz2101 (commented):
The changes made here look good, comment isn't related to this issue but the  `3_reporting` directory in general too, the `results.yaml` in this workflow runs `Sys.date()` as the snapshot date because that's the default set in `BindResults`, wondering if we should make it more "clear" / bring it forward to the `results.yaml` as well

image

samussiah (approved):
Makes sense @lauramaxwell!

lauramaxwell:
> The changes made here look good, comment isn't related to this issue but the `3_reporting` directory in general too, the `results.yaml` in this workflow runs `Sys.date()` as the snapshot date because that's the default set in `BindResults`, wondering if we should make it more "clear" / bring it forward to the `results.yaml` as well
> 
> image

added an [issue](https://github.com/Gilead-BioStats/gsm/issues/1934) to revisit defaults on some of these util functions

PR #1931, opened 2024-11-05
Closes #1882 adds a toggle to display past year or full data

### LINKED ISSUE #1882: Feature: Prevent horizontal scrollbar in Flags Over Time table.
## Feature Details
Display full width of table so user doens't have to scroll to the bottom to scroll to the right.

## Possible Implementation
Or freeze the first three columns and add a scroll bar at the top, but that seems difficult.

## Additional Comments
image

zdz2101:
## Overview
Create multiple table/gt objects in `Widget_FlagOverTime()` to then feed in to the JS script
Modify JS script toggle between the two objects 

Reprex:
```
devtools::install_github("Gilead-Biostats/gsm", ref = "fix-1882")
library(dplyr)
library(gsm)

reportingResultsSubset <- dplyr::filter(
  reportingResults,
  GroupID %in% head(unique(reportingResults$GroupID))
)
reportingResultsSubset2 <- reportingResultsSubset %>%
  mutate(SnapshotDate = SnapshotDate + 365)

reportingResultsSubset3 <- bind_rows(
  reportingResultsSubset2,
  reportingResultsSubset
)

Widget_FlagOverTime(
  dfResults = reportingResultsSubset3,
  dfMetrics = reportingMetrics
)
```

image
image


R/Widget_FlagOverTime.R
```diff
@@ -24,30 +24,51 @@ Widget_FlagOverTime <- function(
...
     "dfResults is not a data.frame" = is.data.frame(dfResults),
     "dfMetrics is not a data.frame" = is.data.frame(dfMetrics),
     "strGroupLevel is not a character" = is.character(strGroupLevel)
   )
 
+  most_recent12 <- dfResults %>%
+    pull(SnapshotDate) %>%
+    unique() %>%
+    sort(decreasing = TRUE) %>%
+    head(12) %>%
```
zdz2101, line 41: Curious what if snapshots are collected biweekly, quarterly, different frequencies, is there a more "flexible" way to be robust for the user?

lauramaxwell: in general, we collect monthly snapshots, but even if it were collected weekly, this would provide roughly a quarter-long view, so i think its a reasonable duration to show as the default

inst/htmlwidgets/Widget_FlagOverTime.js
```diff
@@ -8,10 +8,34 @@ HTMLWidgets.widget({
...
     // Apply styles to make the content scrollable if it gets too long
     el.style.overflowY = 'auto';  // Enable vertical scrolling
+    el.style.overflowX = 'hidden';  // Prevent horizontal scrolling
     el.style.maxHeight = '400px'; // Set the max height for the widget
+    el.style.width = '100%';        // Set width to 100% to prevent overflow
 
     return {
       renderValue: function(x) {
+        // Create the toggle button
+        const toggleButton = document.createElement('button');
```
lauramaxwell, line 18: maybe try to implement the toggle switch, like we use in Report_StudyInfo.R (lines 37-44)

lauramaxwell:
Not sure exactly what the issue is, but the dropdown filter menu is no longer filtering the table properly

lauramaxwell:
Was still seeing an issue with the subset dropdown. part of it had to do with hex code changes in rbm-viz, but haven't totally fixed it. see most recent commit to see what I mean. I'm also ok with pushing this to another PR if it's out of scope here.

lauramaxwell:
Figured out the last piece to be able to close #1935 via js. I think everything is looking good now, but @zdz2101 take a look!

inst/htmlwidgets/Widget_FlagOverTime.js
zdz2101: I think I tried to go somewhere along those lines of just doing it solely in R but I couldn't get past the fact the js widget at its current state only takes 1 table to display, not sure how it would work otherwise @samussiah any thoughts?

lauramaxwell: yeah, at the moment its just a default styled button, but it works as expected. elsewhere this toggle button is inserted via html, so i'm not clever enough to quickly figure out how to do that here in js. we can add another issue if folks are annoyed by how it looks.

lauramaxwell: i actually was able to get like 80% of the way there- got the switch to appear with the label next to it, but it doesn't slide when clicked. i think it's probably worth figuring this out tomorrow before merging this in. @jonthegeek or @samussiah - any tips? 
image

lauramaxwell: got it to slide when clicked!! i'll let someone else take a look before i approve though

inst/htmlwidgets/Widget_FlagOverTime.js
```diff
@@ -8,31 +8,78 @@ HTMLWidgets.widget({
...
           el.insertBefore(footnote, el.firstChild);
         }
 
-        // Insert the flag overtime table content.
-        el.insertAdjacentHTML('beforeend', x.html);
+        // Toggle button to switch between recent (12-month) and full view
+        toggleContainer.addEventListener('click', () => {
+          showRecent = !showRecent;
+          updateTable(showRecent);
+          for(toggle of document.getElementsByTagName('input')){if(toggle.id==='toggle-checkbox')toggle.checked=showRecent;}
```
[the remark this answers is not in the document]
samussiah: This should do the same thing.  Love the toggle!

lauramaxwell (approved):
looking good! since i made edits, i won't merge, but i'm done fiddling with the toggle and i think this is ready to go!

zdz2101:
> looking good! since i made edits, i won't merge, but i'm done fiddling with the toggle and i think this is ready to go!

Certainly improved from before, works for me

samussiah (approved):
Nice work!

PR #1939, opened 2024-11-08
Closes #1933 introduces logging capabilities to gsm

### LINKED ISSUE #1933: QC: add more robust logging using `log4r`
## QC Details

update logging for gsm workflows with `log4r` package so that we can more easily debug with logs both in console and saved out during automated workflow runs

zdz2101:
## Overview
implement `log4r`

## Connected Issues

- Closes #XXX


gsm_log.log
```diff
@@ -0,0 +1,21 @@
+INFO  [2024-11-08 14:07:39] [RunStep] Starting function execution.
```
zdz2101, line 1: @samussiah @lauramaxwell the log here is made by running the testthat file

R/RunStep.R
```diff
@@ -55,10 +55,15 @@
 #' @export
 
 RunStep <- function(lStep, lData, lMeta, lSpec = NULL) {
+  log4r::info(.package_logger, "[RunStep] Starting function execution.")
   # prepare parameter list inputs
   params <- lStep$params
 
   cli::cli_h3("Evaluating {length(params)} parameter(s) for {.fn {lStep$name}}")
+  log4r::info(
```
zdz2101, line 63: It also seems like we can write a function that will run the cli and log4r to get the console message and the log entry, could clean up our cli usage to:

cli::cli_inform ~= log4r::info
cli::cli_warn ~= log4r::warn
cli::cli_abort ~= log4r::error

zdz2101:
@lauramaxwell @samussiah @dpastoor 

Easiest way to demo what I've done is swapping out lines 48-63 of `inst/examples/3_ReportingWorkflow.R` with

```
# Step 1 - Create Mapped Data Layer - filter, aggregate and join raw data to create mapped data layer
mappings_wf <- MakeWorkflowList(strPath = "workflow/1_mappings")
set_logger("mapping")
mapped <- RunWorkflows(mappings_wf, lRaw, strLogger = "mapping")

# Step 2 - Create Metrics - calculate metrics using mapped data
metrics_wf <- MakeWorkflowList(strPath = "workflow/2_metrics")
set_logger("metrics")
analyzed <- RunWorkflows(metrics_wf, mapped, strLogger = "metrics")

# Step 3 - Create Reporting Layer - create reports using metrics data
reporting_wf <- MakeWorkflowList(strPath = "workflow/3_reporting")
set_logger("reporting")
reporting <- RunWorkflows(reporting_wf,
                          c(mapped, list(lAnalyzed = analyzed,
                                         lWorkflows = metrics_wf)),
                          , strLogger = "reporting")

# Step 4 - Create KRI Reports - create KRI report using reporting data
module_wf <- MakeWorkflowList(strPath = "workflow/4_modules")
set_logger("modules")
lReports <- RunWorkflows(module_wf, reporting, strLogger = "modules")
```

The code above will create 4 different logs. You can set the log, plug it into the `strLogger` argument and it will print the log toa file denoted as "xxx.log" where "xxx" is the name of put into `set_logger()` . By default, on package load, the package already sets a default and will automatically log everything to a file called `gsm_log.log` too. 

To get additional information from a user session while they ran workflows, we can save a list of metadata objects when "setting" the logger and retrieve it with `get_logger()` as well

R/util-cli_logger_wrappers.R
```diff
@@ -0,0 +1,55 @@
+#' cli style console appender for gsm
+#'
+#' @param level warning level that maps to log4r
+#' @param ... should contain message and cli_detail
+#'
+#' @export
+cli_fmt <- function(level, ...) {
```
zdz2101, line 7: @dpastoor After reading and understanding what you demonstrated, I started working on the rough draft of the "custom" console_appender layout function that mimics gsm's current cli messaging suite. Swapping over to default_log_layout(), simple_log_layout(), bare_log_layout(), logfmt_log_layout(), json_log_layout(), seems to work quite smoothly already using SetLogger() with the appropriate logger object.

R/util-cli_logger_wrappers.R
```diff
@@ -0,0 +1,55 @@
...
+}
+
+#' Custom logging function that wraps cli messaging
+#' @param level logger levels
+#' @param message message to display; may contain glue-style placeholders
+#' @param cli_detail for cli style alerts the detail for info
+#' @param .envir the environment for glue expressions
+#'
+#' @export
+LogMessage <- function(level, message, cli_detail = NULL, .envir = parent.frame()) {
```
zdz2101, line 38: @lauramaxwell ^ before I proliferate the use of this function for all our `cli` calls, we essentially now have a `logger` object as part of our package with a "custom layout" that essentially wraps/calls `cli`, but if a user/gismo wants a different logging methodology, it can use that instead by swapping in a different `logger` object using `SetLogger()`

lauramaxwell: ah, okay. i think this makes sense- so this just formats the message in the same way we currently display, but `LogMessage()` can be swapped out easily if we want.

zdz2101: To "see" it,

Load this version of the package, try running Step 1, the mapping workflow from example 3.1.

Then run:
```
SetLogger(logger = logger(appender = console_appender(layout = json_log_layout())))
```

and run step 1 example 3.1 again

zdz2101:
@lauramaxwell not quite understanding how the tests would pass through the R-CMD-checks but not the test-coverage one

lauramaxwell:
> @lauramaxwell not quite understanding how the tests would pass through the R-CMD-checks but not the test-coverage one, unable to recreate locally
> 
> P.S. I think I fixed it

Maybe the secret to solving it was just sending this message 😄. Keeping my fingers crossed!

zdz2101:
> > @lauramaxwell not quite understanding how the tests would pass through the R-CMD-checks but not the test-coverage one, unable to recreate locally
> > P.S. I think I fixed it
> 
> Maybe the secret to solving it was just sending this message 😄. Keeping my fingers crossed!

the handling of `{?s}` in a glue message inconsistent so I just wrote it in as `(s)`

R/util-checkSpec.R
```diff
@@ -93,10 +104,21 @@ CheckSpec <- function(lData, lSpec) {
     }
   }
   if (length(missingCols) > 0) {
-    cli::cli_alert_danger("Not all specified columns in the spec are present in the data, missing columns are: {missingCols}")
+    LogMessage(
+      level = "warn",
+      message = "Not all specified columns in the spec are present in the data, missing columns are: {missingCols}"
+    )
   } else if (length(allCols) > 0) {
-    cli::cli_alert("All {length(allCols)} specified column{?s} in the spec are present in the data: {allCols}")
```
zdz2101, line 98: @lauramaxwell specifically this `{?s}` was causing the trouble, otherwise I think everything else is ready for review again

zdz2101:
@dpastoor feel free to take another look

dpastoor (approved):
I think overall you did a great job - and the concept of combining the leveled logging with the cli_alert was a great pattern to keep the interactive UX for a user really top notch.

The one piece of feedback I have is that the way the messages are structured does lose some of the value proposition of the structured workflow option. I don't want this PR to get held up and I'm in favor of merging it in, however it'd be interesting to explore going one level further to actually see if we can get both the structured messages + the CLI-style messages with one pattern without it being too convoluted.

As an example:

```
 LogMessage(
      level = "info",
      message = "SQL Query complete: {nrow(result)} rows returned.",
      cli_detail = "alert_success"
    )
```

from a structured logging perspective, a more "ideal" setup would instead be to have

```
 LogMessage(
      level = "info",
      message = "SQL Query complete",
      nrows = nrow(result),
      cli_detail = "alert_success"
    )
```

such that the resulting log message has a very clear nrows=123 key/value pair. Doing this across _all_ messages means that a parser can reasonably easily parse every single message and turn that into data in systems that need to aggregate results over time.

These messages instead are meant for human consumption and are much harder to extract each individual log message (but easier to read).

I have an idea of how this could work in the future so we can talk about that maybe next week to brainstorm but I think we should merge this in and get this in use now!

jonthegeek:
You can make it play nice with {cli} if we a) separate out any CLI-only bits of the message, and b) take advantage of the `.envir` param in cli:
```r
cli::cli_alert_info("This message has var: {n}.", .envir = list(n = 23))
#> cli::cli_alert_info("This message has var: {n}.", .envir = list(n = 23))
```

So for the example in @dpastoor's comment, maybe something like this:
```r
 LogMessage(
      level = "info",
      message = "SQL Query complete",
      nrows = nrow(result),
      cli_detail = c("alert_success", "{nrows} rows returned.")
    )
```
And then add the `...` of `cli_detail` to the `cli::cli_{cli_detail[[1]]}` call, with `.enviro = {dots of the LogMessage() call}`.

dpastoor:
@jonthegeek exactly! thats pretty much what i was thinking!

zdz2101:
> @jonthegeek exactly! thats pretty much what i was thinking!

@lauramaxwell @jwildfire @samussiah what're your thoughts? shall we merge this in for now and create a new issue for said improvements? Or just make those modifications while this branch is up and alive?

lauramaxwell:
> > @jonthegeek exactly! thats pretty much what i was thinking!
> 
> @lauramaxwell @jwildfire @samussiah what're your thoughts? shall we merge this in for now and create a new issue for said improvements? Or just make those modifications while this branch is up and alive?

@zdz2101 I say we merge it in and add an issue for these modifications for v2.3

PR #1943, opened 2024-11-12
Fix 1801

### LINKED ISSUE #1801: Feature: engine arg in RunQuery
## Feature Details
Add an `engine` argument to `RunQuery()`. Default to an option, then "sqldf" if the option isn't set. Other option is "duckdb" right now. Check that the engine-package is installed for that option (and move sqldf to Suggests, along with duckdb).

## Additional Comments
Be sure to test that both outputs are the same.

annezheng2:
## Overview
Replaces sqldf usage with duckdb for sql processing for RunQuery

## Connected Issues
- Closes #1801


lauramaxwell (approved):
updated `ApplySpec()` to remove the `name__class` format/method that sqldf was using. Dates are passing through and retaining format without trouble, which is great! This is ready to go.

PR #1947, opened 2024-11-14
Release v2.1.2

lauramaxwell:
## Overview

This patch release primarily addresses minor report updates and handling classes in sql queries properly. Specifically:

- Add method param to `RunQuery` so that we can use `"name__class"` method
- update Flags Over Time widget to only show past 12 months by default, but add toggle to show full history
- display site/participant target as `n/N (x%)` in reports
- display study nickname on report (if present)
- use `gt` for tables throughout entire package

## Connected Issues

- Closes #XXX


samussiah (approved):
Another one down!

PR #1949, opened 2024-11-14
attempt to fix qual report pdf

lauramaxwell:
## Connected Issues

- Closes #XXX


samussiah (approved):
Way to go, nice catch!

PR #1953, opened 2024-11-19
Closes #1937 Display only results of metrics where data is available in reports

### LINKED ISSUE #1937: Bugfix:  When metrics don't have any data, don't print the header in the report
## Feature Details

Right now, we always print the headers for all metrics that exist in the `2_metrics` folder. IF these aren't all being created, than the headers will still exist in the report. We should not print out the headers for metrics that we aren't running/don't have data for.

zdz2101:
## Overview
Try "removing" kri0001 and cou0001 from the analysis data in example 3.1

```
# Step 3 - Create Reporting Layer - create reports using metrics data
reporting_wf <- MakeWorkflowList(strPath = "workflow/3_reporting")
reporting <- RunWorkflows(reporting_wf, c(mapped, list(lAnalyzed = analyzed[-c(1,13)],
                                                       lWorkflows = metrics_wf)))
```

### DIFF
--- a/inst/report/Report_KRI.Rmd
+++ b/inst/report/Report_KRI.Rmd
@@ -126,7 +126,7 @@ Report_OverviewText(
 
 ```{r, echo=FALSE, results='asis'}
 
-for (i in unique(params$dfMetrics$MetricID)) {
+for (i in unique(params$dfResults$MetricID)) {
   lMetric <- params$dfMetrics %>% dplyr::filter(MetricID == i) %>% as.list
 
   print(htmltools::h3(lMetric$Metric))


inst/report/Report_KRI.Rmd
```diff
@@ -126,7 +126,7 @@ Report_OverviewText(
 
 ```{r, echo=FALSE, results='asis'}
 
-for (i in unique(params$dfMetrics$MetricID)) {
+for (i in unique(params$dfResults$MetricID)) {
```
zdz2101, line 129: I think this really is the simplest way to fix it @lauramaxwell @samussiah , other "checks" may include making sure the metrics dataset coming out of `3_reporting/metrics.yaml` only has metric IDs of the results?

lauramaxwell: oh nice- i think this makes sense! way easier than what i was thinking we'd have to do 🎉

zdz2101: Unless we want to put in additional guardrails, I think this is actually it

samussiah: Makes sense to me!

lauramaxwell (approved):
looking good!

PR #1977, opened 2024-12-05
Closes #1972 Begin migration of mapping related stuff

### LINKED ISSUE #1972: Feature: Migrate Mapping components to gsm.mapping
## Feature Details
Move all mapping related workflows, functions, tests, vignettes, etc. to gsm.mapping

## Possible Implementation
To ensure it works harmoniously, remove from a branch of gsm where it's stripped, use `remotes` to test functionality

zdz2101:
## Connected Issues

- Closes #XXX


R/RunWorkflow.R
```diff
@@ -159,7 +159,7 @@ RunWorkflow <- function(
   if ("spec" %in% names(lWorkflow)) {
     LogMessage(level = "info", message = "Checking data against spec", cli_detail = "h3")
     # TODO: verify domain names in [ lData ] exist in [ lWorkflow$spec ]
-    CheckSpec(lData, lWorkflow$spec)
+    gsm.mapping::CheckSpec(lData, lWorkflow$spec)
```
zdz2101, line 162: @lauramaxwell `RunWorkflow()` seems to have a `CheckSpec()` call, I'm not sure we want gsm to depend on gsm.mapping? What''re your thoughts?

lauramaxwell: i think this `gsm.utils` package @jonthegeek is talking about is how i was envisioning `gsm.core`, which would not have any "gsm-verse" dependencies, and will most likely be what is left over in gsm after we pull out the mapping and kri stuff. So, because of this, i don't think `CheckSpec()` should be moved over to gsm.mapping because its ok for gsm.mapping to depend on gsm (or gsm.core/gsm.utils in the future), but not the other way around.

lauramaxwell: we can also potential reconsider if we need to checkspec each time a workflow is run... maybe we really only need to checkspec upon ingestion.

jonthegeek: But until gsm.kri is pulled out, gsm.mapping will depend on gsm (for CheckSpec), and gsm will depend on gsm.mapping (to set things up for KRIs). I don't think that will work. We might need to make gsm.utils or gsm.core or whatever before we finish the subdivision, and then gsm can either die or become the "install all gsm packages" package (equivalent to tidyverse) once it's empty.

PR #1986, opened 2024-12-10
fix issue in `RunQuery` related to incomplete column spec

samussiah:
## Overview
An issue popped up running the `Groups` workflow where `RunQuery` couldn't find a `type` attribute on a column spec.  This PR enforces the column spec data structure.

## Test Notes/Sample Code
```
  # Create a sample data frame
  df <- data.frame(
    Name = c("John", "Jane", "Bob")
  )
  lColumnMapping <- list(
    emaN = list(
      source = 'Name'
    )
  )

  # Define the query and mapping
  query <- "SELECT Name as emaN FROM df WHERE Name LIKE '%o%'"

  # Call the RunQuery function and expect no error
  expect_no_error(result <- RunQuery(query, df, bUseSchema = T, lColumnMapping = lColumnMapping))
  expect_equal(class(result$emaN), "character")
```

## Connected Issues

- Closes #XXX

### DIFF
--- a/R/RunQuery.R
+++ b/R/RunQuery.R
@@ -42,18 +42,21 @@ RunQuery <- function(strQuery, df, bUseSchema = FALSE, lColumnMapping = NULL) {
     stop("if use_schema = TRUE, you must provide lColumnMapping spec")
   }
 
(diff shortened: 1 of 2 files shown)


lauramaxwell (approved):
looks good- thanks!

Issue #1987, opened 2024-12-11
allow namespaced function calls in yaml steps


dpastoor:
## Feature Details

Allow RunWorkflows to execute code from other packages

jonthegeek:
I have [a function in the latest gsm.app PR, `util_AsFunction()`](https://github.com/Gilead-BioStats/gsm.app/pull/337/files#diff-3b581349c64cac4370094ad505f8362b66ff06844a1426be16a2a05ac7dec844), that sorts out converting namespaced functions-as-strings to their corresponding functions. I needed it for the plugin system. We could probably move that up to gsm (eventually gsm.core or whatever) if we have multiple use-cases.

dpastoor:
yep thats pretty much exactly what this now is. I'll take a deeper look - we were just using getFromNamespace so i'm not sure if we need the whole pkg_env stuff, but might grab that instead after i look to see what it does real quick

jonthegeek:
@dpastoor Mine is possibly over-flexible, but the idea is for it to work with/without namespace and also if you pass in the bare function.

dpastoor:
yep - we had also debated about the whole giving some syntax to express loading an entire package via attachNamespace as you have as well, where the packages could be defined in the workflow meta section but decided that for now we'll prefer explicitness and reduce likelihood for potential subtle namespace clashes across workflow files until we have tangible cases where its too annoying to namespace the function calls.

I'm going to update th example code once I get the PR in so we have a little more detail on the design for if we ever need to revisit the convo to make it more/less flexible

dpastoor:
existing behavior:

using an existing test that works:

![image](https://github.com/user-attachments/assets/dc2d59fa-fd34-4f2b-8a3c-76f19cca49ee)

same pattern using a namespaced function:

![image](https://github.com/user-attachments/assets/6815d2e0-4020-49d5-96f5-6664215df735)

now works!

![image](https://github.com/user-attachments/assets/a3826bcc-b0ea-4325-99c3-93dd587eeabb)

and we see dplyr doesn't get attached:

![image](https://github.com/user-attachments/assets/cccec77e-1b24-4c91-afd1-80bccd322321)

jonthegeek:
Yeah, I was a bit loose with that because I needed to attach them anyway to make rsconnect see them. I'm sure we can refine the process. That's one reason it'll be good to have this somewhere central!

lauramaxwell:
closed via #1988

PR #1988, opened 2024-12-11
add namespacing capability

dpastoor:
related to discussion in #1987 

This will now allow workflows like:

```
steps:
  - output: lReport
    name: gsm.study.profile::Report_Study_Profile
    params:
      dfResults: Reporting_Results
      dfMetrics: Reporting_Metrics
      dfGroups: Reporting_Groups
```

to properly find the function calls they need.


R/util-func-namespace.R
```diff
@@ -0,0 +1,15 @@
...
+  fn_pieces <- strsplit(strFunction, "::")[[1]]
+  if (length(fn_pieces) > 2) {
+    rlang::abort("function pieces should only be namespaced via <pkg>::<fn>, detected more :: calls")
+  }
+  if (length(fn_pieces) == 2) {
+    pkg <- fn_pieces[[1]]
+    fn <- fn_pieces[[2]]
+    return(rlang::as_function(fn, env = getNamespace(pkg)))
+  }
+  return(strFunction)
```
jonthegeek, line 14:
```suggestion
  return(rlang::as_function(strFunction))
```
Oh yeah, I forgot you can use functions as strings in almost all situations (see `"mean"(1:10)`). Still, it probably wouldn't hurt to `rlang::as_function` this, to make sure it works as a function everywhere.

R/util-func-namespace.R
```diff
@@ -0,0 +1,15 @@
+##' getFunctionIfNamespaced looks to see if a strFunction is namespaced and looks it up
+##' allowing the do.call in run step to process correctly
+##' this will either return a function, or the original string if not namespaced
```
jonthegeek, line 3: Can we export this? Once we have it nailed down, I'll use it in gsm.app for plugin functions. I'd genericize the name a little, too, since this function will parse strFunction whether or not it's namespaced.

tests/testthat/test-RunStep.R
```diff
@@ -45,3 +45,12 @@ test_that("Handles multiple parameters and function invocation correctly", {
...
 })
+
+test_that("RunStep will run a function from a namespace", {
+  lStep <- list(name = "dplyr::glimpse", params = list(head(Theoph)))
+  lData <- list(data1 = 300)
+  lMeta <- list(meta1 = 400)
+
+  result <- RunStep(lStep, lData, lMeta)
+  expect_equal(result, head(Theoph))
+})
```
jonthegeek, line 56: Add a test for a non-namespaced function to make sure it doesn't ever break anything.

jonthegeek (commented):
I'm requesting changes, but not flagging it as "Request changes" because it's fine to merge this as-is. I'd just love it if we could genericize this a bit so we have one place to edit this, rather than a near-duplicate function in gsm.app.

R/util-func-namespace.R
dpastoor: i'd be completely fine with that, the reason i didn't do it was for expediency and I didn't know if there could be any reason that rlang::as_function could cause a behavioral shift compared to passing do.call a string. I waffled on adding that and figured I'd go conservative for time. If your experience is this also works the same then i'm all for it

lauramaxwell (approved):
LGTM! added a few edits to _pkgdown.yml and NEWS.md to prep for the v2.2.0 RC

tests/testthat/test-RunStep.R
lauramaxwell: test added!

PR #1994, opened 2024-12-12
Closes #1976 wrap all our stop-esque calls into the log4r setup

### LINKED ISSUE #1976: QC: replace `stop()` with `rlang::abort()` throughout package
## QC Details

use `rlang::abort()` instead of `stop()` for fatal errors so that it can provide more information, rather than just the message. See [this](https://www.onceupondata.com/post/handling-r-errors/) for more information on the differences

zdz2101:
## Connected Issues

- Closes #XXX


R/util-MakeBounds.R
```diff
@@ -31,9 +31,19 @@ MakeBounds <- function(
   dSnapshotDate = unique(dfResults$SnapshotDate),
   strStudyID = unique(dfResults$StudyID)
 ) {
-  stop_if_empty(strMetrics)
-  stop_if_empty(dSnapshotDate)
-  stop_if_empty(strStudyID)
+  # stop_if(cnd = !length(strMetrics), message = "`strMetrics` must not be `NULL`.")
+  # stop_if(cnd = !length(dSnapshotDate), message = "`dSnapshotDate` must not be `NULL`.")
+  # stop_if(cnd = !length(strStudyID), message = "`strStudyID` must not be `NULL`.")
```
zdz2101, line 36: @lauramaxwell @jonthegeek @samussiah @jwildfire which looks better ? The single line call for each stop statement or a big chunked one demo'd below

This would affect all the `stopifnot()` calls too, and realistically all the `if()` then `stop()` calls can be wrapped already with commits here can be wrapped in too

R/util-MakeBounds.R
```diff
@@ -31,9 +31,19 @@ MakeBounds <- function(
...
+
+  stop_if_any(
+    conditions = exprs(!length(strMetrics), !length(dSnapshotDate), !length(strStudyID)),
```
lauramaxwell, line 39: looks like we need to add `rlang exprs` to the dependencies. was getting some errors when running workflows and tests.

zdz2101: I want to say this will only be necessary for if we go with this `stop_if_any()` that accepts all the conditions & messages at once, the single call one I dont think its necessary as I use `enquo()` which was already in the `utils-tidy-eval.R`

R/util-Report.R
```diff
@@ -6,15 +6,31 @@
...
 }
 
-stop_if_empty <- function(x, x_arg = rlang::caller_arg(x)) {
-  if (!length(x)) {
-    LogMessage(
-      level = "fatal",
-      message = "{x_arg} must not be `NULL`."
-    )
+stop_if <- function(cnd, message) {
+  # Use enquo to capture the condition for lazy evaluation
```
samussiah, line 10: What's the point of `enquo` and `eval_tidy`-ing the logical expression?  Is there a scenario where a data mask is passed into `stop_if`?

R/util-MakeBounds.R
samussiah: One statement per condition feels simpler, and `stop_if` is more intuitive than `stopifnot`.

R/util-Report.R
zdz2101: Think when I was initially drafting toy examples I needed to evaluate objects that may not necessarily have been initialized/don't exist yet, so it was necessary there, but now that I think about it, I suppose all the conditions we feed into cnd already are TRUE/FALSE. Let me see if this simplication applies and just spread it everywhere.

samussiah: Only other thought was maybe you want to pass the un-evaluated logical statement to the logger. Either way TIL about `enquo` and `eval_tidy` - cool stuff!

lauramaxwell (approved):
a few super minor, picky edits to improve readability of some of the conditions. Other than that- this looks great!

PR #1998, opened 2024-12-17
Fix #1992- always return function in namespace util

### LINKED ISSUE #1992: Feature: Always Convert strFunction to function
## Feature Details
`GetStrFunctionIfNamespaced(strFunction)` returns a function if `strFunction` is namespaced, and a character vector if not. This works in gsm's context, but, in gsm.app, I'm expecting a function, and use it as such after running the string-to-function parser. I let this through in #1988, but it'd be nice if we always `rlang::as_function()` at the end to make sure the thing is a function. It isn't a HUGE issue (I can easily work around it), but it'd be nice.

## Possible Implementation
`rlang::as_function(<something that's already a function>)` doesn't hurt anything, so running that at the end of the function is probably enough (that's effectively what I'm doing).

lauramaxwell:
## Overview

Simple fix to always have the same output class (function) in the `GetStrFunctionInNamespace()`

## Connected Issues

- Closes #1992

### DIFF
--- a/R/util-func-namespace.R
+++ b/R/util-func-namespace.R
@@ -4,7 +4,7 @@
 #'
 #' This function looks to see if a strFunction is namespaced and looks it up
 #' allowing the do.call in run step to process correctly
-#' this will either return a function, or the original string if not namespaced
+#' this will return a function.
 #'
 #' @param strFunction the function to be called
 #' @export
--- a/R/util-func-namespace.R
+++ b/R/util-func-namespace.R
@@ -3,8 +3,8 @@
 #' `r lifecycle::badge("experimental")`
 #'
 #' This function looks to see if a strFunction is namespaced and looks it up
-#' allowing the do.call in run step to process correctly
-#' this will return a function.
+#' allowing the do.call in run step to process correctly.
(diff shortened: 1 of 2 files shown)


R/util-func-namespace.R
```diff
@@ -21,5 +21,7 @@ GetStrFunctionIfNamespaced <- function(strFunction) {
     fn <- fn_pieces[[2]]
     return(rlang::as_function(fn, env = getNamespace(pkg)))
   }
-  return(strFunction)
+  else {
+    return(rlang::as_function(strFunction))
+  }
```
jonthegeek, line 26:
```suggestion
  return(rlang::as_function(strFunction))
```
Since you've already `return()`ed the other case, you don't even need `else` here.

jonthegeek (changes requested):
There's a small tweak you can do to make it even cleaner. Not 100% *necessary*, but it's nice. Oh, and fix the vignette errors, evidently!

jonthegeek (approved):
Looks like that did it! Pkgdown isn't done yet when I'm typing this, so I hope I don't jynx it!

PR #2006, opened 2024-12-31
Fix #1996- add study-wide average for data entry lag

### LINKED ISSUE #1996: Feature :Add average days to entry KRI
## Feature Details

Add average days to entry to the outstanding data entry rate scatterplot/description. May require update to rbm-viz as well depending on where the information gets added

From Jenn: 
>  It would be really useful to be able to say how poorly the site is doing relative to the study’s data entry expectations. The value could be way over 10 days potentially and it would be meaningful to be able to provide that in the signal description if that were the case.

lauramaxwell:
## Overview

Added some text under the Outstanding Data Entry KRI to specify the study-level average for data entry lag (in days) as well as a study-level average of the data entry lag for those that were flagged as >= 10 days of lag. This is related to a request from Jenn. Will run it by her to confirm this is what she wanted prior to merging in.

## Test Notes/Sample Code

image

## Connected Issues

- Closes #1996


lauramaxwell:
I _think_ i need to add the site-level average(s) to the tooltip so we can compare more directly.

lauramaxwell:
Edited this to be a new metric that uses Sum of data entry lag days as the numerator, and a strType = rate for Analyze_NormalApprox. This is what the output looks like on clindata: 
image
image

lauramaxwell:
not doing this as a KRI- will be a deep dive app plugin most likely

PR #2007, opened 2025-01-06
Review package dependencies

### LINKED ISSUE #1979: QC: Review Imports and Suggests in NAMESPACE and DESCRIPTION
## QC Details
Reduce dependencies as possible

lauramaxwell:
Removes unused dependencies for the gsm package.

Closes #1979


jonthegeek (approved):
Looks good!

One ~style comment: I prefer to only @importFrom (or @import) things in these categories:
- Things that are used all over the place (eg `cli_abort`).
- Things to allow you to put a package in Imports without R CMD Check freaking out (eg sometimes that's necessary for `glue` or `lifecycle`).
- Things that are difficult to namespace (`%||%` from `rlang`, etc).

For everything else, just `pkg::function` when they're used. That way you don't keep `@importFrom`ing after the function is no longer used.

Not a necessary change, just my $0.02!

jonthegeek:
Oops! Looks like the specific method we use might come from {broom} specifically.

PR #2019, opened 2025-01-15
Remove glossary.

jonthegeek:
## Overview
This idea never came together fully, so let's simplify.

You still need `R/aaa-shared.R` in each package that splits off from gsm, but then ideally we should delete any parameters that aren't used from there. Search for `gloss_param` in your repo to find other place where this was used, and copy/paste the equivalent from here.


lauramaxwell (approved):
working, and easier to handle when breaking gsm up. thanks for the quick PR!

PR #2012, opened 2025-01-09
Add button to show/hide report table of contents

lauramaxwell:
## Overview

Per CM team request, adding a hamburger button to allow the user to hide the table of contents in the kri report. 

## Test Notes/Sample Code

We might want to think about adjusting the default width of the report to be a percentage of the screen instead of a fixed width, but not sure if that is needed at the moment.

## Connected Issues

- Closes #1963

PR #2028, opened 2025-01-24
Revert "Add button to show/hide report table of contents"

samussiah:
Reverts Gilead-BioStats/gsm#2012 - I'd like to move the toggle into the ToC itself if possible.  The toggle's overlapping a bit.


lauramaxwell (approved):
yeah, that works for me-i'll approve this and i think  the`fix-1963` branch still exists, so you should be able to just start from there.

PR #2041, opened 2025-03-06
Closes #2036 remove clindata dependency and create datasim object in data-raw/data folder

### LINKED ISSUE #2036: QC: Remove clindata dependency
## QC Details

Use `gsm.datasim` instead of `clindata`. Also, remove the `UseClindata()` function.

zdz2101:
## Overview
swap clindata references with gsm.datasim
put gsm.datasim object in data-raw and data


lauramaxwell (approved):
this looks great! looking at  the report that this produces, we definitely need to work on making more "interesting" data in `gsm.datasim`, but once we make those edits, the data can be updated here.

PR #2045, opened 2025-03-25
gsm deprecation messages

lauramaxwell:
## Overview

Added a notice upon loading in  `.onAttach` as well as at the top of the README to inform the user of the replacement of gsm with a set of modularized packages.

## Connected Issues

- Closes #XXX


jonthegeek (approved):
This is great! This will definitely be helpful to move users to the new packages.

